PR #84, opened 2015-12-12
Fixing a couple issues

rkuchan:
Fixing the following issues:
* Correct the regex claim in `data_exploration.md`. Fixes: https://github.com/influxdb/docs.influxdata.com/issues/51
* Edit `schema_exploration.md` to discuss `SHOW MEASUREMENT` working with regexes. Fixes: https://github.com/influxdb/docs.influxdata.com/issues/46
* Add a FEI section about `ALTER RETENTION` not affecting the current shard. Fixes: https://github.com/influxdb/docs.influxdata.com/issues/48

I'm a bit wary of the third fix, because:
* There was mention of a 30 day shard length here: https://github.com/influxdb/influxdb/issues/4441#issuecomment-151997416
* In that same discussion it mentions that moving from a 1hr RP to an infinite RP drops the current 1hr shard - that confuses me and I may have misread that
* The drawing is huge and a bit unattractive - I just want it to help!

Finally - I moved the FEI "writing more than one CQ to a single series" from the Querying Data section to the Writing data section.


content/influxdb/v0.9/troubleshooting/frequently_encountered_issues.md
```diff
@@ -285,11 +288,21 @@ There a number of ways to identify the version of InfluxDB that you're using:
...
 
+| RP duration  | Shard group interval  |
+|---|---|
+| < 2 days  | 1 hour  |
+| >= 2 days and <= 6 months  | 1 day  |
+| > 6 months  | 7 days  |
 
+*Example:*
 
+Shard groups cover a seven day time interval if you're working with an infinite RP. If you change the `DURATION` of that RP to one hour, shard groups now cover a one hour time interval. When InfluxDB enforces the new, shorter RP, it drops all of the seven-day length shard groups that are older than one hour. If, however, at the time of the RP change, InfluxDB is still writing to a seven day long shard it cannot drop that shard. It must wait until it has finished writing to that shard to drop it.
```
gunnaraasen: > If, however, at the time of the RP change, InfluxDB is still writing to a seven day long shard it cannot drop that shard. It must wait until it has finished writing to that shard to drop it.

Any shard group with an interval that overlaps with the current retention policy duration is kept around, not just shard groups active for writes. This happens because InfluxDB will not automatically split up existing shard groups into new shard groups.

gunnaraasen: I think a `1d` duration would be better for this example.

content/influxdb/v0.9/troubleshooting/frequently_encountered_issues.md
```diff
@@ -285,11 +288,21 @@ There a number of ways to identify the version of InfluxDB that you're using:
...
 
+In the image below, the RP change occurs on day five of the second seven-day shard group. After the RP change, InfluxDB drops the first seven-day shard, but because it's still writing to the second seven-day shard, it must keep that shard for another 48 hours - this is why older data may stick around even if you've recently shortened the `DURATION` of an RP.
```
gunnaraasen: I think labelling the shards with numbers would be helpful when referencing the picture, e.g. "...drops the first seven-day shard, but..." becomes "...drops shard 1, but".

Also, for naming consistency, lets try using "shard groups" when talking about retention policies.

content/influxdb/v0.9/troubleshooting/frequently_encountered_issues.md
```diff
@@ -312,3 +288,24 @@ There a number of ways to identify the version of InfluxDB that you're using:
 * Check the HTTP response in your logs:  
 
 `[http] 2015/09/04 12:29:07 ::1 - - [04/Sep/2015:12:29:06 -0700] GET /query?db=&q=create+database+there_you_go HTTP/1.1 200 40 -` ✨`InfluxDBShell/0.9.3`✨ `357970a0-533b-11e5-8001-000000000000 6.07408ms`
+
+## Experiencing unexpected data retention after altering a retention policy
```
gunnaraasen: "Data not dropped after altering retention policy"

PR #106, opened 2015-12-17
Add discussion on COUNT() with fill()

rkuchan:
I added a section about `fill()` with functions (in general) which links back to the Data Exploration page. 

I also added a section that discusses the specific behavior of `COUNT()` with `fill()`. This fixes https://github.com/influxdb/docs.influxdata.com/issues/96.

### DIFF
--- a/content/influxdb/v0.9/query_language/functions.md
+++ b/content/influxdb/v0.9/query_language/functions.md
@@ -82,7 +82,7 @@ time			               count
 
 > #### `COUNT()` and controlling the values reported for intervals with no data
 > <br>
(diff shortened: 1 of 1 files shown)


beckettsean:
+1

Not sure when we are switching to 0.10 for the docs. @gunnaraasen should we be making those changes now? Seems like we can't launch the docs until late Jan if these changes are in place, or are we documenting 0.10 before its release?

rkuchan:
I changed all the links back to v0.9 and I also added a short section on including multiple functions in a single query. That last change fixes https://github.com/influxdb/docs.influxdata.com/issues/116.

beckettsean:
+1 LGTM

PR #114, opened 2015-12-23
Update shell.md

rkuchan:
Updates the CLI page with the arguments and commands available in 0.9.6.1. Also made some formatting changes and added an in-depth discussion about `-import`.

Fixes: https://github.com/influxdb/docs.influxdata.com/issues/109

### DIFF
--- a/content/influxdb/v0.9/tools/shell.md
+++ b/content/influxdb/v0.9/tools/shell.md
@@ -8,181 +8,139 @@ menu:
     parent: tools
 ---
 
-The Influx shell is an interactive shell for InfluxDB, and is part of all InfluxDB distributions starting with the InfluxDB 0.9.0. The shell is not compatible with InfluxDB 0.8 and earlier. If you install InfluxDB via a package manager, the shell is installed at `/usr/bin/influx` (`/usr/local/bin/influx` on OS X).
+The InfluxDB's command line interface (`influx`) is an interactive shell for the database, and is part of all InfluxDB distributions starting with version 0.9.0. Use `influx` to write data (manually or from a file), query data interactively, and view query output in different formats.
 
-## Shell Arguments
+* [Launch `influx`](/influxdb/v0.9/tools/shell/#launch-influx)
+* [`influx` Arguments](/influxdb/v0.9/tools/shell/#influx-arguments)
+* [`influx` Commands](/influxdb/v0.9/tools/shell/#influx-commands)
 
-There are several arguments you can pass into the shell when starting.  You can list them by passing in `--help` to get the following results:
+## Launch `influx`
+If you [install](https://influxdata.com/downloads/) InfluxDB via a package manager, the CLI is installed at `/usr/bin/influx` (`/usr/local/bin/influx` on OS X).
 
-```sh
-$ influx --help
-Usage of influx:
-  -version
-       Display the version and exit.
-  -host 'host name'
-       Host to connect to.
-  -port 'port #'
-       Port to connect to.
-  -database 'database name'
-       Database to connect to the server.
-  -password 'password'
-      Password to connect to the server.  Leaving blank will prompt for password (--password '')
-  -username 'username'
-       Username to connect to the server.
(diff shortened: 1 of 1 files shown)


content/influxdb/v0.9/tools/shell.md
```diff
@@ -8,181 +8,139 @@ menu:
...
-
-        a full list of influxql commands can be found at:
-        http://influxdb.com/docs
-```
+`-consistency 'any|one|quorum|all'`  
+Set the write consistency level.
 
-### connect
+`-database 'database name'`  
+The database to connect to.
```
beckettsean: The database to which to connect


(I have split infinitives in writing, I know it's pedantic. Ignore me if you think it's stilted or reduces comprehension.)

content/influxdb/v0.9/tools/shell.md
```diff
@@ -8,181 +8,139 @@ menu:
...
+Specifies the format of the server responses. See [-format](/influxdb/v0.9/tools/shell/#specify-the-format-of-the-server-responses-with-format).
 
-You do not need specify both parts of the server.  For example,
-if your current host is `localhost:8086`, the following command:
-
-```sh
-> connect :8087
-```
+`-host 'host name'`  
+Host to connect to. By default, InfluxDB runs on localhost.
```
beckettsean: Host to which to connect.

content/influxdb/v0.9/tools/shell.md
```diff
@@ -8,181 +8,139 @@ menu:
...
-and use those credentials when querying the database.
+`-port 'port #'`  
+Port to connect to. By default, InfluxDB runs on port `8086`.
 
-### settings
+`-pps`  
+How many points per second the import will allow. By default, pps is zero and influx will not throttle importing. Use with `-import`.
 
-Settings will output the current state of the shell.
+`-precision 'h|m|s|ms|u|ns'`
```
beckettsean: I think `rfc3339` is another valid input. It is from within the shell.

content/influxdb/v0.9/tools/shell.md
```diff
@@ -8,181 +8,139 @@ menu:
...
+`-port 'port #'`  
+Port to connect to. By default, InfluxDB runs on port `8086`.
 
-### settings
+`-pps`  
+How many points per second the import will allow. By default, pps is zero and influx will not throttle importing. Use with `-import`.
 
-Settings will output the current state of the shell.
+`-precision 'h|m|s|ms|u|ns'`  
+Specifies the precision of the timestamp: `h` (hours), `m` (minutes), `s` (seconds), `ms` (milliseconds), `u` (microseconds), `ns` (nanoseconds). Precision defaults to nanoseconds.
```
beckettsean: Is this precision specified for writes, queries, or both? I use it to set the output format but I've never noticed if it can chance the precision for `insert` or `import` files.

content/influxdb/v0.9/tools/shell.md
```diff
@@ -204,76 +171,109 @@ Pretty print enabled
...
+`pretty`  
+Turns on pretty print for the `json` format.
+
+`settings`  
+Outputs the current settings for the shell including the `Host`, `Username`, `Database`, `Pretty` status, `Format`, and `Write Consistency`.
+
+`use <db_name>`  
+Sets the current database. Once `influx` sets the current database, there is no need to specify that database in queries. `influx` automatically queries the current database and its `DEFAULT` retention policy.
+
+#### Write data to InfluxDB with `insert`
```
beckettsean: I think `INSERT INTO` can be used to insert into a specific retention policy, but I haven't played with it enough to find out. Since there's no way to set RP in the CLI that would be good to document if it works.

Can you experiment with `INSERT INTO` and see if the behavior makes sense and is something we should document?

rkuchan:
I made the requested edits. I'm a little concerned that the note about `-precision rfc3339` in the `influx Arguments` section is confusing.

PR #117, opened 2015-12-23
update kapacitor docs for new alert handler

nathanielc:
Also change weights for auto vs manual created pages.


beckettsean:
@rossmcdonald @kostasb @gunnaraasen maybe you can give @nathanielc some feedback on the docs changes?

content/kapacitor/v0.2/tick/alert_node.md
```diff
@@ -256,10 +314,12 @@ node.info(value tick.Node)
 ### Log
 
 Log JSON alert data to file. One event per line. 
+Must specify the absolute path the the log file.
```
gunnaraasen: "the the log file" > "of the log file"

content/kapacitor/v0.2/tick/alert_node.md
```diff
@@ -44,6 +46,8 @@ or the alert just changed to the &#39;OK&#39; state from a non &#39;OK&#39; stat
 Using the [AlertNode.StateChangesOnly](/kapacitor/v0.2/tick/alert_node/#statechangesonly) property events will only be sent to handlers 
 if the alert changed state. 
 
+It is valid to configure multiple alert handlers of the same type and of different types.
```
gunnaraasen: > It is valid to configure multiple alert handlers of the same type and of different types. 

How about: "It is valid to configure multiple alert handlers, even with the same type.

content/kapacitor/v0.2/tick/alert_node.md
```diff
@@ -173,6 +168,69 @@ node.flapping(low float64, high float64)
...
+Send alert to HipChat using default room &#39;Test Room&#39;. 
+**NOTE**: The global option for HipChat also implies stateChangesOnly is set on all alerts. 
+Also, the room can either be the room id (numerical) or the room name. 
+
+
+```javascript
+node.hipChat()
+```
+
+Properties of HipChat
```
gunnaraasen: How about changing the headings below to something like

```
#### HipChat Room Property
```

and removing this line.

gunnaraasen:
+1

On a side note, the alerts page will grow larger as more alerts are added. I think it might make sense to break out the alert node and other large node pages (and also the `*Out` node pages) into separate sub-menus.

nathanielc:
I have addressed to issues. I would like to merge this as the last update to the v0.2 docs. 

I will be submitting a v0.10 PR here in a bit. 

The semantic line feed changes make this look ugly. These Kapacitor docs are auto generated from the comments in the Kapacitor code itself. As a result the semantic line feed doesn't really work well. But you don't really edit these files directly either; the process to update them is to update the comments in the code first and then just regen them. Is that going to drive anyone crazy that these file don't follow the semantic line feed concept?

gunnaraasen:
I'm fine with merging this and not using semantic line feeds for Kapacitor docs. It might be stylistically inconsistent, but throwing out the autogen seems like a much bigger waste. There may also be a possibility of adding semantic linefeed support to the autogen library in the future.

@beckettsean Are you ok with foregoing semantic linefeeds in the Kapacitor docs?

beckettsean:
@nathanielc I totally agree with @gunnaraasen. The primary reason to do semantic line feeds is to facilitate manual PR reviews. If the content is automatically generated the need for micro-edits in PRs is greatly reduced and semantic linefeeds are of less value. Given the other challenges it would introduce, I see no overall benefit from them.

nathanielc:
Sounds good. merging...

nathanielc:
@gunnaraasen @beckettsean sounds good, merging....

gunnaraasen:
Thanks @nathanielc! Let me know if you need any help setting up the v0.10 PR.

PR #124, opened 2016-01-04
Functions with GROUP BY time require a WHERE time clause

kostasb:
The select statements using the 'GROUP BY time' aggregate function won't work unless there is a 'WHERE time' clause in the query.
This was not depicted in the queries of this overview.


beckettsean:
+1 merge it

Issue #131, opened 2016-01-06
move to semantic linefeeds


beckettsean:
As per https://github.com/influxdata/rfc/issues/3#issuecomment-169418400

rossmcdonald:
@beckettsean Is that the correct link?

beckettsean:
Not even close. Paste buffer FTL. Edited original comment. Thanks, @rossmcdonald!

rkuchan:
Fixed by https://github.com/influxdata/docs.influxdata.com/pull/149

PR #149, opened 2016-01-14
Switching to semantic linefeeds

beckettsean:
Quick and dirty regex to do most of it. Numbered lists were not well preserved, I don't know of any other issues yet.

I used `s/([\.\!\?])\s+(.*?)/$1\n$2` to create semantic line feeds in the docs. Before that substitution, multiple consecutive `\n` were protected using `s/{\n,}/XXXXX/` and then later restored with `s/XXXXX/\n\n/`.


beckettsean:
@rkuchan I think numbered lists will need some fixing but I haven't seen anything else amiss so far.

beckettsean:
Fixes #131

rkuchan:
Thanks, @beckettsean! I went through and fixed up the numbered lists. 👍

beckettsean:
Here goes nothing...

@gunnaraasen @rkuchan @mjdesa @rossmcdonald @kostasb @jackzampolin @toddboom please keep an eye on the docs after the next push. This was a huge mostly automated change, and I'm sure at least one or two inadvertent edits slipped through.

PR #157, opened 2016-01-15
Dark Theme for Docs

alexpaxton:
## Overview

The previous version of docs I saw as my “first stab” at the design. Just enough to get going. This time around I’ve been able to take my time and really refine the interface. This PR contains a collection of different changes, some huge some small. I took this opportunity to 

**Entirely new things**

- Dark theme (Trying to avoid blinding people switching from code editors)
- Header images (A la Medium and the modern blog aesthetic, brings some pretty pictures in and heightens the experience of reading)
- Header images are selected at random via javascript from a set of 14, after page load. I am anticipating this may be problematic on slow connections, though 90% of our traffic comes from desktop so may be negligible.
- Support advertisement (Hopefully this gets us a lot more sales, may want to run some A/B tests on the copy and/or design of the ad to optimize its effectiveness)

**Adjustments**

- Gutted the old `typography.css` file. Switched to a mathematically calculated type scale using classic ratios. Feels much nicer to read. The entire content area can be scaled proportionally by changing `font-size` on `.page` to a value using `rem` as the unit.
- Standardized the indentation on indented items
- Very subtle image in the background of code blocks. Intended to delight and break the monotony of the text. Hopefully it doesn’t distract.
- Mobile view is a little different because of the changes to the page header
- Styled `.flash` messages to be more in your face

**Bugs, misc.**
- Looks like the version number does not always show up in `.product-switcher--current`
- Are the links in the navbar that point back to influxdata.com the links we still want? Curious to see if these links are effective at sending traffic to the main site, and if that traffic converts or not. If these links aren’t contributing much I’d love to remove them

**Requests**
- Last edited timestamp on each article. @gunnaraasen mentioned this is difficult because files are modified via github and hugo is unaware. I believe this will contribute a lot to the user experience of the Docs site, as it allows users to see how “reliable” each article is. Also gives us transparency in showing how often we update articles
- Previous / Next article links in the header and footer of each article
- Ability to associate a unique image with each article. I hear this is a decent amount of work so I would say low priority, but if anyone wants to make this designer happy…

#### Closing notes:
- If you are reviewing this PR, I am looking for primarily usability feedback rather than aesthetic
- If you can find unusual combinations of markdown output that looks janky


layouts/_default/section.html
```diff
@@ -1,12 +1,11 @@
...
-    <div class="page-title--toggle" id="page-title--toggle"><i class="icon arrow-right"></i></div>
-    <div class="page-title--name">default/section{{ .Title }}</div>
-</div>
-{{ end }}
-<div class="page-content" id="page-content">
-    <div class="page-content--container">
+<div class="page" id="page-content">
+    {{ if .Title }}
+    <div class="page--title" id="page-title">
+        <div class="page--body">default/section{{ .Title }}</div>
```
gunnaraasen: The `default/section` needs to be removed. I think that's an artifact from when I was fiddling with the templates.

layouts/_default/single.html
```diff
@@ -1,12 +1,11 @@
...
-	<div class="page-title--toggle" id="page-title--toggle"><i class="icon arrow-right"></i></div>
-	<div class="page-title--name">{{ .Title }} (Default Template)</div>
-</div>
-{{ end }}
-<div class="page-content" id="page-content">
-	<div class="page-content--container">
+<div class="page" id="page-content">
+	{{ if .Title }}
+	<div class="page--title" id="page-title">
+		<div class="page--body">{{ .Title }} (Default Template)</div>
```
gunnaraasen: Same for `(Default Template)` here.

alexpaxton:
@gunnaraasen should I remove the extraneous template files?

gunnaraasen:
Yeah, the extraneous template files can be removed except the layouts in the `_default` directory, since those are the fallback templates.

layouts/influxdb/section.html
```diff
@@ -1,14 +1,14 @@
 {{ partial "header.html" . }}
-<div class="page-title" id="page-title">
-    <div class="page-title--toggle" id="page-title--toggle"><i class="icon arrow-right"></i></div>
-    <div class="page-title--name">{{ .Title }} Section</div>
-</div>
-<div class="page-content" id="page-content">
-    <div class="page-content--container">
+<div class="page" id="page-content">
+    <div class="page--title" id="page-title">
+	    <div class="page--body">{{ .Title }} Section</div>
```
gunnaraasen: Same for `Section` here.

gunnaraasen:
I'm fine with killing the links pointing back to influxdata.com, but I'm sure @jginfluxdata @ShubhraKar will want to weigh in on that decision.

gunnaraasen:
+1

@alexpaxton I have another idea on a way to pull the last page edited timestamp. I'll give that a go, but don't hold your breath. I'll also work on getting the next/prev added as well.

beckettsean:
Merging on your behalf, @alexpaxton. We're excited!

PR #168, opened 2016-01-20
kapacitor v0.10 docs

nathanielc:
I think I got it all right. At least locally the version and sidebar all seem to be working.

Note the only additions from v0.2 are the UDF and Stats nodes.

 I plan to write up an example on how to use UDFs but that will be a different PR.


nathanielc:
@gunnaraasen I clicked merge on the wrong tab :( so sorry. I can push a new PR if you have any changes that need to get in. If it broke something feel free to backout the change I can easily regenerate the changes.

gunnaraasen:
No worries. There are a couple files named `content/kapacitor/v0.10/tick/.*.md.swp` that look like they snuck in. Everything else looks good.

PR #199, opened 2016-02-04
Minor updates

### LINKED ISSUE #196: [0.10.0] Port 8091 seems to be undocumented
Hello,

it seems like port 8091 seems to be missing from documentation at https://docs.influxdata.com/influxdb/v0.10/concepts/api/

Thanks

### LINKED ISSUE #197: Remove reporting section from InfluxDB configuration overview
This section:

https://docs.influxdata.com/influxdb/v0.10/administration/config/#reporting

Makes it seem like there should be a `[reporting]` section in the configuration, which is not present in any of the example configurations. It should be updated to show that `reporting-disabled` is a global variable.

Brought up by https://github.com/influxdata/influxdb/issues/5528

rossmcdonald:
Two changes here:

- Added port 8091 to port section. Fixes #196.

- Added a Global Options section to the configuration (removing Reporting). Fixes #197.

### DIFF
--- a/content/influxdb/v0.10/administration/config.md
+++ b/content/influxdb/v0.10/administration/config.md
@@ -36,7 +36,7 @@ See the [installation documentation](/influxdb/v0.10/introduction/installation/#
 
 ## Configuration Sections
 
-* [[reporting]](/influxdb/v0.10/administration/config/#reporting)
+* [Global Options](/influxdb/v0.10/administration/config/#global-options)
 * [[meta]](/influxdb/v0.10/administration/config/#meta)
 * [[data]](/influxdb/v0.10/administration/config/#data)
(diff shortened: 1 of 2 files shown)


rossmcdonald:
@gunnaraasen Regarding the second change above, do you know if the `[reporting]` section of the config documentation should be there? I've changed it to 'Global Options' since there was some confusion about why the docs differed from the config sample. I've never seen a `[reporting]` section before, but I wanted to verify before removing it.

rkuchan:
@rossmcdonald, that was me actually. I wanted to have a header to link to in the docs when I listed out each section at the top - I see that that could be confusing so feel free to remove it.

rossmcdonald:
@rkuchan Ah, understood. I think it makes more sense to just have a 'global options' section, since there are going to be a few more variables added there in the next release (I think).

rkuchan:
Sounds good to me!

PR #201, opened 2016-02-04
Add useful commands to Server Monotoring page

jackzampolin:
@gunnaraasen @rossmcdonald @kostasb

### DIFF
--- a/content/influxdb/v0.10/administration/statistics.md
+++ b/content/influxdb/v0.10/administration/statistics.md
@@ -8,7 +8,7 @@ menu:
 
 InfluxDB can display statistical and diagnostic information about each node.
 This information can be very useful for troubleshooting and performance monitoring.
-
+ 
 ## SHOW STATS
 To see node stats execute the command `SHOW STATS`.
 An example is shown below.
@@ -61,3 +61,20 @@ This returns information such as build information, uptime, hostname, server con
 InfluxDB also writes statistical and diagnostic information to database named `_internal`, which records metrics on the internal runtime and service performance.
 The `_internal` database can be queried and manipulated like any other InfluxDB database.
 Check out the [monitor service README](https://github.com/influxdb/influxdb/blob/master/monitor/README.md) and the [internal monitoring blog post](https://influxdb.com/blog/2015/09/22/monitoring_internal_show_stats.html) for more detail.
+
+## Useful Commands
+
+Below are a collection of commands to find useful performance metrics about your InfluxDB instance. 
+
+To find the number of unique series in a database run the following command: 
+```bash
+$ influx -execute "SHOW SERIES" -database "mydb" | sed '/name: .*/d' | sed '/\-\-/d' | sed '/_key/d' | sed '/^$/d' | wc -l
+```
+To find the number of points per second being written to the instance:
+```bash
+$ influx -execute 'select derivative(pointReq, 1s) from "write" where time > now() - 5m' -database '_internal' -precision 'rfc3339'
+```
+To find the number of writes separated by database:
+```bash
+$ grep 'POST' /var/log/influxdb/influxd.log | awk '{ print $10 }' | sort | uniq -c 
+```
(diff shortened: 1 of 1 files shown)


content/influxdb/v0.10/administration/statistics.md
```diff
@@ -61,3 +61,20 @@ This returns information such as build information, uptime, hostname, server con
...
+```bash
+$ influx -execute "SHOW SERIES" -database "mydb" | sed '/name: .*/d' | sed '/\-\-/d' | sed '/_key/d' | sed '/^$/d' | wc -l
+```
+To find the number of points per second being written to the instance:
+```bash
+$ influx -execute 'select derivative(pointReq, 1s) from "write" where time > now() - 5m' -database '_internal' -precision 'rfc3339'
+```
+To find the number of writes separated by database:
+```bash
+$ grep 'POST' /var/log/influxdb/influxd.log | awk '{ print $10 }' | sort | uniq -c
```
rossmcdonald: You might want to mention that this will only show how many writes per database since the log was rotated/created.

content/influxdb/v0.10/administration/statistics.md
```diff
@@ -61,3 +61,20 @@ This returns information such as build information, uptime, hostname, server con
...
+
+Below are a collection of commands to find useful performance metrics about your InfluxDB instance. 
+
+To find the number of unique series in a database run the following command: 
+```bash
+$ influx -execute "SHOW SERIES" -database "mydb" | sed '/name: .*/d' | sed '/\-\-/d' | sed '/_key/d' | sed '/^$/d' | wc -l
+```
+To find the number of points per second being written to the instance:
+```bash
+$ influx -execute 'select derivative(pointReq, 1s) from "write" where time > now() - 5m' -database '_internal' -precision 'rfc3339'
```
rossmcdonald: I'm not sure about this one. On my local instance (which has historical data, but nothing feeding into it right now), I see:

```
influx -execute 'select derivative(pointReq, 1s) from "write" where time > now() - 5m' -database '_internal' -precision 'rfc3339'
name: write
-----------
time			derivative
2016-02-04T21:02:58Z	4.1
2016-02-04T21:03:29Z	-320111.3548387097
2016-02-04T21:03:39Z	4.4
2016-02-04T21:03:49Z	3.1
2016-02-04T21:03:59Z	4.5
2016-02-04T21:04:09Z	3.1
2016-02-04T21:04:19Z	4.5
2016-02-04T21:04:29Z	3.8
2016-02-04T21:04:39Z	3.8
2016-02-04T21:04:49Z	3.1
2016-02-04T21:04:59Z	4.5
2016-02-04T21:05:09Z	3.8
2016-02-04T21:05:19Z	3.1
2016-02-04T21:05:29Z	104.5
2016-02-04T21:05:39Z	3.2
2016-02-04T21:05:49Z	4.6
2016-02-04T21:05:59Z	3.9
2016-02-04T21:06:09Z	3.9
2016-02-04T21:06:19Z	3.2
2016-02-04T21:06:29Z	4.6
2016-02-04T21:06:39Z	3.9
2016-02-04T21:06:49Z	3.9
2016-02-04T21:06:59Z	3.2
2016-02-04T21:07:09Z	4.6
2016-02-04T21:07:19Z	3.2
2016-02-04T21:07:29Z	4.6
2016-02-04T21:07:39Z	3.2
```

Maybe this shows points from the `monitor` service as well? We may want to check with a Core developer to make sure these numbers mean what we think they mean.

rossmcdonald: Actually the large negative number may have been a retention check, so this may be correct. I'll monitor the output of this on a few different databases to verify.

rossmcdonald: We may also want to mention that the `monitor` plugin has to be enabled for this command to work.

rossmcdonald:
Two comments, but looks great other than that. :+1:

content/influxdb/v0.10/administration/statistics.md
jackzampolin: It does show points from the `monitor` service as well. I've checked this against the numbers that `influx_stress` outputs and other derived `pps` measurements and it always lines up very closely. Also works for clusters if you add `group by hostname`.

rossmcdonald: Okay, awesome, that's good to know. We should probably also add a note about querying across a cluster.

content/influxdb/v0.10/administration/statistics.md
gunnaraasen: Agree this needs a disclaimer.

gunnaraasen:
+1 this looks good. Ready to merge once the disclaimer on the last command is added.

PR #203, opened 2016-02-04
Fix indentation for one in consistency

ajgb:
"one" was at the same level as its parent "consistency"


gunnaraasen:
Thanks @ajgb

PR #208, opened 2016-02-05
Added disclaimer to clustering documentation

rossmcdonald:
The disclaimer was accidentally dropped with the 0.10.0 migration. It needs to be put back.


gunnaraasen:
+1

Issue #220, opened 2016-02-09
Add page describing how InfluxDB clustering works


rossmcdonald:
Currently there are no in-depth details discussing how clustering works under the hood (how shards are distributed, how distributed queries are run, etc).

Issue #218, opened 2016-02-08
Entries to add to the glossary


rkuchan:
* [x] points per second
* [x] values per second
* [x] shard
* [x] shard group
* [x] wal
* [x] tsm
* [x] metastore
* [x] consensus node
* [x] hybrid node
* [x] consensus service
* [x] data service
* [x] line protocol

beckettsean:
fixed by #485

PR #223, opened 2016-02-09
Edit clustering guide

rkuchan:
I made a lot of formatting changes and tried to fill out some of the descriptions.


content/influxdb/v0.10/guides/clustering.md
```diff
@@ -9,83 +9,211 @@ menu:
...
-activity in the cluster.  Meta nodes do not require significant system
-resources and can run on a very lightweight server.
+## InfluxDB cluster model
 
-- [Data nodes](/influxdb/v0.10/concepts/glossary/#data-node) store data
-and respond to queries. Data nodes must run on systems with at least
-2 CPUs, 4GB RAM, and storage with 1000 IOPS.  See the
-[hardware sizing guide](/influxdb/v0.10/guides/hardware_sizing/) for
-more detail.
+InfluxDB supports arbitrarily sized clusters and any [replication factor](/influxdb/v0.10/concepts/glossary/#replication-factor) from 1 to the number of nodes in the cluster.
```
gunnaraasen: It seems like replication factors could use some additional info here, or the mention of replication could be moved further down the page.

content/influxdb/v0.10/guides/clustering.md
```diff
@@ -9,83 +9,211 @@ menu:
...
+There are three types of nodes in an InfluxDB cluster:
```
gunnaraasen: Listing out the meta-data node as a separate type doesn't feel right. I think it might be better to frame this as each InfluxDB process (whether running on a different hosts or not) can have different behaviors in a cluster. I'm not really happy with that wording either.

content/influxdb/v0.10/guides/clustering.md
```diff
@@ -9,83 +9,211 @@ menu:
...
 
-Any node can be a meta node, data node, or both. Each cluster must
-have _at least_ three meta nodes in order to form a Raft concensus and
-remain in a healthy state.
+* **Meta nodes** coordinate activity in the cluster. They participate in consensus, which means that they have consistent data about cluster membership, [databases](/influxdb/v0.10/concepts/glossary/#database), [retention policies](/influxdb/v0.10/concepts/glossary/#retention-policy-rp), [users](/influxdb/v0.10/concepts/glossary/#user), [continuous queries](/influxdb/v0.10/concepts/glossary/#continuous-query-cq), and shard meta-data.
```
gunnaraasen: cluster membership -> servers in a cluster

gunnaraasen: Link to the definition of shard in the glossary.

content/influxdb/v0.10/guides/clustering.md
```diff
@@ -9,83 +9,211 @@ menu:
...
 
-## Configuration
+    *Requirements:*
+    Meta nodes do not require significant system resources and can run on a very lightweight server.
 
-The following is the current recommended procedure for configuring a cluster.
+* **Data nodes** store data and respond to queries. Here, data refers to time series data and the time series data [schema](/influxdb/v0.10/concepts/glossary/#schema).
```
gunnaraasen: "Data nodes store data and respond to queries." -> "Data nodes store the actual time series data and respond to queries."

Nix the mention of the time series data schema.

content/influxdb/v0.10/guides/clustering.md
```diff
@@ -9,83 +9,211 @@ menu:
...
 
-> **Note:** You should always use the most recent release for clustering as there are significant improvements with each release.
+    *Requirements:*
+    Data nodes must run on systems with at least 2 CPUs, 4GB RAM, and storage with 1000 IOPS.
```
gunnaraasen: Lets remove this reference and keep all the hardware recommendations in the hardware sizing guide. It'll be easier to maintain changing recommendations on one page in the future.

content/influxdb/v0.10/guides/clustering.md
```diff
@@ -9,83 +9,211 @@ menu:
...
-_three_ nodes in order to form a healthy raft cluster. It is also
-assumed that you are running some version of Linux, and, while it is
-possible to build a cluster locally, it is not recommended.
+In an InfluxDB cluster, any node can be a meta node, data node, or both.
+Each cluster must have _at least_ three meta nodes (or meta-data nodes) in order to form a Raft consensus and remain in a healthy state.
 
-1. Install InfluxDB on the 3 machines following the [installation guide](/influxdb/v0.10/introduction/installation/).
+## Cluster setup
+
+The following sections configure an InfluxDB cluster with three meta nodes (the raft cluster) and one data node.
```
gunnaraasen: Why this configuration? I think the more common use case would be a four server cluster with three meta-data nodes and one data-only node.

content/influxdb/v0.10/guides/clustering.md
gunnaraasen: "Meta-data" sounds too close to "metadata", which we use elsewhere to describe the data stored in a meta node. It would be great to have some better terminology for these behaviors. The current terms are confusing.

content/influxdb/v0.10/guides/clustering.md
```diff
@@ -9,83 +9,211 @@ menu:
...
 
-At this point you'll want to verify that your initial raft cluster is healthy.
-To do this, issue a `SHOW SERVERS` query to each node in your raft cluster using the `influx` CLI.
-You should see something along the lines of this:
+[http]
+  ...
+  bind-address = "IP:8086"
+```
+
+● Setting `[meta] enabled = true` and `[data] enabled = false` makes the node a meta node.
```
gunnaraasen: Use a `-` or `*` for lists instead of `●`.

content/influxdb/v0.10/guides/clustering.md
```diff
@@ -9,83 +9,211 @@ menu:
...
+id	 http_addr  tcp_addr
+1	  IP1:8091	  IP1:8088
+2	  IP2:8091	  IP2:8088
+4	  IP3:8091	  IP3:8088
 ```
 
+> **Note:** The irregular node `id` numbers is a known issue and a fix is underway.
+For now, it may be easier to identify data nodes and meta nodes by the IP addresses reported in the `SHOW SERVERS` results.
+
+And that's the Raft cluster!
```
gunnaraasen: I think we might want to avoid direct references to raft. People may not be familiar with it and InfluxDB abstracts out direct interaction with the consensus protocol.

gunnaraasen:
This is a good start! I think we should come up with some better terminology for the different pieces of a cluster.

content/influxdb/v0.10/guides/clustering.md
```diff
@@ -9,83 +9,211 @@ menu:
...
 name: data_nodes
 ----------------
-id	http_addr	tcp_addr
-2	influx1:8086	influx1:8088
-5	influx3:8086	influx3:8088
-6	influx2:8086	influx2:8088
+id	 http_addr	 tcp_addr
+1	  IP1:8086	  IP1:8088
+3	  IP2:8086	  IP2:8088
+5	  IP3:8086	  IP3:8088
```
rossmcdonald: If the `[data]` configuration is [disabled](https://github.com/influxdata/docs.influxdata.com/pull/223/files#diff-946cd094a0026aec8c814279d83448d1R63), then I don't think any nodes should display under the `data_nodes` section.

rossmcdonald:
I'm worried we're overcomplicating things by recommending people split out meta nodes and data nodes like this. Users are confused enough by our clustering configuration as it is, and I think splitting them out in the configuration is just going to lead to more confusion.

Is there a reason why we're not just recommending that they run 3 meta/data nodes? That seems like the simplest and most reliable configuration.

content/influxdb/v0.10/guides/clustering.md
beckettsean: I think it's okay this way, but that might be because "replication factor" sounds intuitive to me by now.

content/influxdb/v0.10/guides/clustering.md
beckettsean: What about calling it a "Dual node"? That way we have Data node, Meta node, and Dual node.

That doesn't scale if we add more types of nodes in the future, though. 

I think we should kick off a naming brainstorm for this. I'll start an email.

content/influxdb/v0.10/guides/clustering.md
beckettsean: I'm okay with cluster membership, but again, I might just be unable to detect our internal jargon anymore.

content/influxdb/v0.10/guides/clustering.md
beckettsean: Agreed, although I do want to keep qualitative statements.

* Data nodes must run on server hardware appropriate to your data needs. See the hardware sizing guide...

content/influxdb/v0.10/guides/clustering.md
beckettsean: I do think we should have at least one example of setting up a meta/data node in the cluster, but I also like the idea of showing a meta only node setup.

Perhaps we combine the two suggestions into a four node cluster with 1 meta, 1 data, and 2 meta-data nodes?

gunnaraasen:
I agree with @rossmcdonald on this. I think it might be time to have a clustering guide and separate clustering docs with a description on all the various clustering options. Then people can guide themselves after they go through the guide.

rossmcdonald:
I like the idea of having two different guides/docs:
- An introduction to clustering that covers creating the base configuration of 3 meta/data nodes
- An advanced clustering guide that covers creating more complex configurations, where meta and data nodes are split out

Also we've never really tested on the configurations that we're recommending, which will most likely lead to more support issues further down the road.

content/influxdb/v0.10/guides/clustering.md
rkuchan: Agreed on the shard link - it's not currently in the glossary so I created [this issue](https://github.com/influxdata/docs.influxdata.com/issues/218).

content/influxdb/v0.10/guides/clustering.md
rkuchan: Agreed - I was having trouble with this in markdown - it kept reading my `*`s and `-`s as numbers. I'll take another look at this.

content/influxdb/v0.10/guides/clustering.md
rkuchan: My bad. I set up meta-data nodes while writing this 😬

rkuchan:
I agree on reducing the level of complication here. Because this is currently the only doc on clustering I wanted to include each kind of configuration. 

I like the idea of separating out the complex configurations into a new doc. I'm leaning towards giving clustering its own section - it'll have the two docs that Ross mentioned above as well as a doc on how clustering actually works (https://github.com/influxdata/docs.influxdata.com/issues/220).

rkuchan:
@gunnaraasen, @rossmcdonald and @beckettsean, thank you for all the feedback! Same goes to everyone who responded to Sean's email about names.

The last commit dedicates a section of the docs to clustering. Clustering has two docs:
* `cluster_setup.md` - this walks though how to set up a three hybrid node cluster. I removed a lot of the detail that appeared in the original version to make it more guide-like - there's just a short section at the end that discusses adding more nodes. I also removed the hardware guidelines and added a section about consensus nodes to `hardware_sizing.md`.
* `cluster_node_config.md` - this shows how to configure consensus nodes, data nodes, and hybrid nodes.

The last commit also updates the 0.10 docs for the new naming system (so meta node is now consensus node, meta-data node is now hybrid node, and data node is still data node):
* `glossary.md` now has an entry for hybrid node
* `09_vs_010.md` has updated language about clustering

One things to note is that in `cluster_setup.md` the results from `SHOW SERVERS` splits results into `meta_nodes` and `data_nodes`. This will probably be confusing. Also - having a doc on how clustering works would be awesome.

rkuchan:
@beckettsean, thank you! 

I updated the docs with all of your edits but one; I didn't add an example of `influx -host` to `cluster_setup.md`. There are a couple notes around `SHOW SERVERS` already and didn't want to overwhelm anyone. Happy to add it in if you feel strongly about it.

Todos:
* Link to the still-to-come issue about renaming meta nodes to consensus nodes in the `SHOW SERVERS` output.
* Maybe edit the warning at the top of the page based on input from Fred and Jason.

content/influxdb/v0.10/concepts/09_vs_010.md
```diff
@@ -49,9 +49,10 @@ InfluxData recommends [converting](https://github.com/influxdata/influxdb/blob/m
 
 ## Clustering
 
-InfluxDB 0.9 allowed data-only nodes. InfluxDB 0.10 allows data-only nodes as well as meta-only nodes.
-This feature allows users to separate out meta nodes in very large clusters or configurations where data nodes are under heavy load. Another option is to have two nodes that act as meta and data nodes and one less-powered node that acts as a meta-only node.
+InfluxDB 0.9 allowed data-only nodes. InfluxDB 0.10 allows [data-only nodes](/influxdb/v0.10/concepts/glossary/#data-node) as well as [consensus nodes](/influxdb/v0.10/concepts/glossary/#consensus-node).
```
rossmcdonald: Should this be:

> InfluxDB 0.9 supported only hybrid nodes.

Since they were consensus/raft and data rolled into one?

rossmcdonald:
+1, with one comment

content/influxdb/v0.10/concepts/09_vs_010.md
rkuchan: Thank you for catching that @rossmcdonald.

beckettsean: Actually, 0.9 supported hybrid or data-only. There was no consensus-only, but it was possible to run a data-only node.

content/influxdb/v0.10/concepts/glossary.md
```diff
@@ -19,6 +19,27 @@ See [Clustering](/influxdb/v0.10/guides/clustering/) for how to set up an Influx
...
+
+## consensus service
+The InfluxDB service that participates in the raft consensus group.
+A cluster must have at least three nodes running the consensus service (consensus or hybrid nodes), but it can have more.
+There should be an odd number of nodes running the consensus service in a cluster.
+
+The number of consensus services that can fail before the cluster is degraded is &#8968;n/2 + 1&#8969; where `n` is the number of consensus services in the cluster.
+Thus, an even number of consensus services offer no additional redundancy or resiliency.
+
+The consensus service has consistent data about cluster membership, databases, retention policies, users, continuous queries, shard metadata, and subscriptions.
```
beckettsean: The consensus service ensures consistency across the cluster for node membership, databases, ...

beckettsean:
:+1: with one wording change and one correction (0.9 allowed data-only nodes)

PR #230, opened 2016-02-11
Replacing occurences of ```shell with  ```sh

kyzh:
Change all the code highlighting from `shell` to `sh` so it is consistent across the documentation


kyzh:
closing this PR as it is not correct see https://github.com/influxdata/docs.influxdata.com/issues/232#issuecomment-183129651

PR #231, opened 2016-02-11
Replacing occurences of ```sh with  ```shell

kyzh:
Change all the code highlighting from `sh` to `shell` so it is consistent across the documentation


kyzh:
closing this PR as it is not correct see https://github.com/influxdata/docs.influxdata.com/issues/232#issuecomment-183129651

Issue #232, opened 2016-02-11
standardize on either sh or shell for shell code highlight


kyzh:
I'm looking at the documentation and for highlighting shell code, there is a mix of
```
```sh
```
and 
```
```shell
```

I created 2 PR:
- PR#230 to standardize on [sh](https://github.com/influxdata/docs.influxdata.com/pull/230)
- PR#231 to standardize on [shell](https://github.com/influxdata/docs.influxdata.com/pull/231)

That way you can just pick your favourite.

gunnaraasen:
Thanks @kyzh. Looking at the Prism.js docs realized both `sh` and `shell` are incorrect. It should be `bash`.

kyzh:
I'll submit a PR with `bash` instead then.

kyzh:
@gunnaraasen I closed the two original PR and open a new one with the suggested fix [here](https://github.com/influxdata/docs.influxdata.com/pull/237)
Thanks for :mag: at my PRs

gunnaraasen:
Fixed by https://github.com/influxdata/docs.influxdata.com/pull/237

PR #236, opened 2016-02-12
Replacing occurences of ```sh and ```shell  

kyzh:
As per https://github.com/influxdata/docs.influxdata.com/issues/232#issuecomment-183129651 the correct value is 
```
```bash
```


kyzh:
My sed was not correct :disappointed:
I'll close this one and re-open a PR, this time with the correct fix

PR #238, opened 2016-02-12
making the disclaimer less scary

beckettsean:
updating 0.10 to be more descriptive and less offputting.

### DIFF
--- a/content/influxdb/v0.10/guides/clustering.md
+++ b/content/influxdb/v0.10/guides/clustering.md
@@ -7,7 +7,7 @@ menu:
     parent: guides
 ---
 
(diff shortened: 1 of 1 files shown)


beckettsean:
@rkuchan @mjdesa @rossmcdonald @jackzampolin @jwilder @jginfluxdata @fred-influx 

Are you okay with this update to the clustering guide?

jwilder:
No.  I would say it is still experimental in 0.10.

gunnaraasen:
@beckettsean the proposed wording makes clustering sound better than it is. I think we can remove the "experimental" label but still make the warning more descriptive of the current state of clustering.

E.g. "Clustering works but is still in active development and not yet feature complete. There are rough edges and edge-case failure scenarios have not been fully tested. If you encounter any issues, please [report them](https://github.com/influxdata/influxdb/issues/new)."

beckettsean:
Leaving it as-is for now, will update when a more nuanced picture is available.

PR #243, opened 2016-02-14
Remove hostname from links

lenn4rd:
While browsing the InfluxDB documentation, I found a link that had `localhost:1313` in it. I fixed three occurrences after searching within the entire directory.


rkuchan:
Thank you for finding those, @lenn4rd!

lenn4rd:
Don’t mention it! :)

PR #249, opened 2016-02-16
Add search plus some other design updates

gunnaraasen:


gunnaraasen:
@rkuchan @beckettsean @kostasb Is this good to merge?

beckettsean:
:+1: 

We need to make a blog post about it, and mention:

"Currently the lookahead feature of the search box returns 0.9 and 0.10 results independently, which leads to seemingly duplicate results. We are working with the Algolia, the search service provider, to improve the experience. We hope that even an imperfect search tool is better than none, and we welcome your feedback."

or something like that.

Issue #244, opened 2016-02-15
Service - UDP docs typo


red97corvette:
In the config example for the High-Traffic UDP scenario the UDP port number is 80891, which is invalid (max 65535)

https://docs.influxdata.com/influxdb/v0.10/write_protocols/udp/

rkuchan:
Thanks, @red97corvette for pointing that out!

Issue #216, opened 2016-02-06
Precision for UDP?


tpitale:
This page suggests precision can be set in the "query string" https://docs.influxdata.com/influxdb/v0.10/write_protocols/line/. How does it get set when using UDP? What is the default precision for UDP?

beckettsean:
@sparrc @jwilder is it possible to control the timestamp precision when writing over UDP, or is it always nanosecond precision?

jwilder:
It's nanosecond, but @tpitale submitted (and we merged yesterday) a PR to allow changing the precision via the config.  https://github.com/influxdata/influxdb/pull/5565

beckettsean:
Awesome!

@rkuchan we need to update the UDP docs when influxdata/influxdb#5565 merges

jwilder:
@beckettsean @rkuchan  It's already merged as of yesterday.

PR #250, opened 2016-02-16
UPD updates

rkuchan:
Fix UDP port and add `precision` option to UDP config.

Fixes https://github.com/influxdata/docs.influxdata.com/issues/244, https://github.com/influxdata/docs.influxdata.com/issues/216


gunnaraasen:
:+1:

Issue #221, opened 2016-02-09
search 0.10 docs for overwriting a point


beckettsean:
Related to https://github.com/influxdata/influxdb/issues/1647#issuecomment-180642599

In 0.10 it is no longer possible to overwrite the field set for a point. We have used that in the past as a delete workaround for dropping individual points, but it no longer works.

beckettsean:
Presumably the right solution is to use `DELETE SERIES` with a very tight `WHERE time` clause, but that isn't yet wired up. https://github.com/influxdata/influxdb/issues/1647

PR #255, opened 2016-02-16
Update docs for 010 overwrite behavior

rkuchan:
In 0.10 it is no longer possible to overwrite the field set for a point. This updates the glossary entry for point, the FEI that discusses writing duplicate points, and adds a note to `09_vs_010.md`. It also removes the FEI that talks about writing more than one CQ to a single series. 

Issues:
* The section in `09_vs_010.md` is still rough. Should I include a note about this being a workaround for deletes in 0.9 and now, with 0.10, users must wait for `DELETE SERIES` to perform deletes?

Fixes https://github.com/influxdata/docs.influxdata.com/issues/221

### DIFF
--- a/content/influxdb/v0.10/troubleshooting/frequently_encountered_issues.md
+++ b/content/influxdb/v0.10/troubleshooting/frequently_encountered_issues.md
@@ -32,7 +32,6 @@ Where applicable, it links to outstanding issues on GitHub.
 * [Writing integers](/influxdb/v0.10/troubleshooting/frequently_encountered_issues/#writing-integers)   
 * [Writing duplicate points](/influxdb/v0.10/troubleshooting/frequently_encountered_issues/#writing-duplicate-points)  
 * [Getting an unexpected error when sending data over the HTTP API](/influxdb/v0.10/troubleshooting/frequently_encountered_issues/#getting-an-unexpected-error-when-sending-data-over-the-http-api)
-* [Writing more than one continuous query to a single series](/influxdb/v0.10/troubleshooting/frequently_encountered_issues/#writing-more-than-one-continuous-query-to-a-single-series)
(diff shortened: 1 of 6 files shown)


content/influxdb/v0.10/troubleshooting/frequently_encountered_issues.md
```diff
@@ -269,27 +268,57 @@ Writes an integer: `value=100i`
...
-For example,
-<br>`cpu_load,hostname=server02,az=us_west value=24.5 1234567890000000` and
-<br>`cpu_load,hostname=server02,az=us_west value=5.24 1234567890000000` are identical points.
-The last one written will overwrite the other.
+For example:
 
-In order to store both points, simply introduce an arbitrary new tag to enforce uniqueness.
-<br>`cpu_load,hostname=server02,az=us_west,uniq=1 value=24.5 1234567890000000` and
-<br>`cpu_load,hostname=server02,az=us_west,uniq=2 value=5.24 1234567890000000` are now unique points, and each will persist in the database.
+Old point: `cpu_load,hostname=server02,az=us_west val=24.5,feld=7 1234567890000000`
```
gunnaraasen: feld -> field

gunnaraasen: Nevermind, looks like `feld` was intentional?

rkuchan: That was intentional - probably not a great field name though. I can change that.

gunnaraasen:
I agree there should be a note about the old delete workaround in 0.9 no longer works in `09_vs_010.md`. Also, there should probably be a note that `DELETE SERIES` does not work in 0.10 in the FEI too with a link to the GH issue.

gunnaraasen:
:shipit:

rkuchan:
Thank you, @gunnaraasen!

Issue #265, opened 2016-02-23
CQ field, tag warning


rkuchan:
InfluxDB 0.10

https://docs.influxdata.com/influxdb/v0.10/query_language/continuous_queries/

Add add a warning about: A tag in the `SELECT` clause of a CQ statement turns into a field in the new measurement. To preserve tags put them in the `GROUP BY` clause. Including a tag in the `GROUP BY` clause and the `SELECT` clause will make the data unquery-able. Something like that.

PR #267, opened 2016-02-23
Add note about field,tag behavior

rkuchan:
Adds a note about how to preserve tags with CQs and warns about including a tag key in the `SELECT` clause and `GROUP BY` clause. Fixes https://github.com/influxdata/docs.influxdata.com/issues/265.


gunnaraasen:
+1

PR #273, opened 2016-02-24
Bugfixes for a few issues

### LINKED ISSUE #254: Update CREATE DATABASE documentation for deprecated IF NOT EXISTS feature
In 0.10, the `IF NOT EXISTS` syntax is now ignored when creating a database and it is the default and only behavior. We will be deprecating the syntax in 0.12 (from influxdata/influxdb#5707), but we need to include a deprecation notice and update the documentation for the change in behavior so confusions like influxdata/influxdb#5563 don't happen.

### LINKED ISSUE #257: With outdated pages, have the top 0.10 link redirect to the equivalent doc, not the main 0.10 docs page
It's relatively common for me to:

* Google something like `SHOW SERVERS influxdb`
* Click the top result which is still for 0.9
* Click to go to the most recent version of that docs page(0.10)
* End up at the main docs page where I have to find my desired page again.

### LINKED ISSUE #268: Update LAST() documentation
What products and version are you using?
InfluxDB 0.10

URL for relevant page?
https://docs.influxdata.com/influxdb/v0.10/query_language/functions/

Where did you look before opening the issue?
https://github.com/influxdata/influxdb/issues/5796#issuecomment-187882596

---

Explain that `LAST()` returns a single point when there is no `GROUP BY` clause and two points have the same timestamp. There is no order as to the point it picks to return.

Also - review `FIRST()` documentation as that might do the same thing.

### LINKED ISSUE #270: Warning about CQs with same name
Version: 0.10
Doc: https://docs.influxdata.com/influxdb/v0.10/query_language/continuous_queries/

Only one CQ runs if you have more than one CQ with the same name in different databases. This should be fixed in 0.11.
See https://github.com/influxdata/influxdb/issues/5814

### LINKED ISSUE #272: Typo in Administration - Config Documentation
> This section controls where the **actural** shard data for InfluxDB lives and how it is flushed from the WAL. dir may need to be changed to a suitable place for you system, but the WAL settings are an advanced configuration. The defaults should work for most systems.

should be

> This section controls where the **actual** shard data for InfluxDB lives and how it is flushed from the WAL. dir may need to be changed to a suitable place for you system, but the WAL settings are an advanced configuration. The defaults should work for most systems.

What products and version are you using?
InfluxDB v0.10 docs

URL for relevant page?
https://docs.influxdata.com/influxdb/v0.10/administration/config/#data
https://github.com/influxdata/docs.influxdata.com/blob/master/content/influxdb/v0.10/administration/config.md#data

Where did you look before opening the issue?

gunnaraasen:
Fixes #272 
Fixes #270 
Fixes #268 
Fixes #257 
Fixes #254 

@rkuchan

### DIFF
--- a/content/influxdb/v0.10/administration/config.md
+++ b/content/influxdb/v0.10/administration/config.md
@@ -152,7 +152,7 @@ Meta logging toggles the logging of messages from the meta service.
 
 ## [data]
 
-This section controls where the actural shard data for InfluxDB lives and how it is flushed from the WAL. `dir` may need to be changed to a suitable place for you system, but the WAL settings are an advanced configuration. The defaults should work for most systems.
+This section controls where the actual shard data for InfluxDB lives and how it is flushed from the WAL. `dir` may need to be changed to a suitable place for you system, but the WAL settings are an advanced configuration. The defaults should work for most systems.
 
 ### dir = "/var/opt/influxdb/data"
 
--- a/content/influxdb/v0.10/query_language/continuous_queries.md
+++ b/content/influxdb/v0.10/query_language/continuous_queries.md
@@ -47,6 +47,9 @@ CREATE CONTINUOUS QUERY ON <database_name> [RESAMPLE [EVERY <interval>] [FOR <in
 A CQ belongs to a database.
 Specify the database where you want the CQ to live with `ON <database_name>`.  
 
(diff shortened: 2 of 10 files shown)


rkuchan:
Five issues! SO HAPPY. Everything looks great - just one comment.

rkuchan:
👍

PR #279, opened 2016-02-26
Harmonisation of Kapacitor introduction with InfluxDB introduction

kyzh:
This change include:

- Adding content to the "Introduction" page (when one click the left panel menu "Introduction").
- Adding a link to the Download page on the left panel under "Introduction" sub menu
- Swapping the getting started and installation order, so you install Kapacitor before using it
- Adding a more comprehensive download procedure by adding md5sum check before installation step
- Making the top index under Kapacitor the same format as the other

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -1495,3 +1495,11 @@ SectionPagesMenu = "products"
   weight = 100
   url = "https://github.com/CargoSense/puppet-influxdb"
   parent = "tools"
+
+# Kapacitor v0.10 Sub-Menu - Introduction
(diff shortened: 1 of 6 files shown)


beckettsean:
@kyzh Thanks very much for the work you did here! We hope to improve on the current InfluxDB Getting Started and are in the process of completely rethinking the Getting Started flow. To keep changes to a minimum we are leaving the other product docs as-is until we have a new Getting Started structure on InfluxDB.

PR #282, opened 2016-02-29
collectd Config Updates

rossmcdonald:
Updated collectd documentation with a little more information, and a link to the service readme.

https://github.com/influxdata/influxdb/issues/5836
https://github.com/influxdata/influxdb/issues/5867


gunnaraasen:
+1

PR #285, opened 2016-03-03
Remove erroneous spaces in bash command

speakingcode:
The spaces in the environment variable assignments ($VERSION_ID = "8" and $VERSION_ID = "7") break the commands to add the InfluxData repositories in Debian systems. Removing the spaces yields correct commands.


beckettsean:
@rossmcdonald

rossmcdonald:
@speakingcode Good catch. Thanks, @speakingcode!

+1

PR #298, opened 2016-03-09
highlight multiple fields per point

beckettsean:


rkuchan:
The bulleted list of changes at the top of the page is missing `Multiple fields per point is now more efficient`. Otherwise 👍

beckettsean:
Thanks, fixed!

PR #307, opened 2016-03-11
Small typo.

rubycut:


rkuchan:
Good catch, @rubycut. Thank you for fixing!

PR #310, opened 2016-03-14
v0.11 kapacitor docs

nathanielc:
New v0.11 Kapcitor docs.

Small changes to content as things were changed from v0.10 to v0.11.
Biggest change is the addition of an Index section to all Node docs for which methods are propery vs chain methods at glance.

TODO:
- [x] There are several breaking change from v0.10 to v0.11 and the upgrade path has been noted in the CHANGELOG. We should write an explicit section in the docs on upgrading so it is easier to find.


nathanielc:
@rossmcdonald Here are the new docs for v0.11 of Kapacitor. We should coordinate our PRs so we don't duplicate work. 

I think we should let your PR go through first and update the v0.10 docs and then I'll update this PR to match.

nathanielc:
@rossmcdonald This is ready to go, it now mirrors the same restructure as your PR.

rossmcdonald:
@nathanielc I added a few commits to clean up formatting and grammar. Let me know if you'd like to discuss any of them.

For the InfluxQL node, I think that we should highlight the new function calls in a more prominent way. They are listed under the "Chaining Methods" section, but they get mingled with all of the pre-existing chain calls (like alert, deadman, etc.). What if we put the InfluxQL calls at the top of the list, instead of alphabetically listed? I imagine they are auto-generated, so I'm not sure what our options are here. Thoughts?

Also, please see https://github.com/influxdata/kapacitor/pull/343. Does that change look okay? They look much better if they're highlighted in the doc styling (using the `>`), but it's not super important.

nathanielc:
> For the InfluxQL node, I think that we should highlight the new function calls in a more prominent way. They are listed under the "Chaining Methods" section, but they get mingled with all of the pre-existing chain calls (like alert, deadman, etc.). What if we put the InfluxQL calls at the top of the list, instead of alphabetically listed? I imagine they are auto-generated, so I'm not sure what our options are here. Thoughts?

They are auto-generated but we can make it do what we need. Maybe a section for just the InfluxQL functions? Not sure what to do the reality is there is much of a difference programatically between `alert` and `count` they are both methods that perform their respective action. I don't feel the strong need to group them but I understand the sentiment.

rossmcdonald:
I like the idea of a section at the top. I understand that they're programmatically the same, but from an end-user perspective the reason I'd probably go to the InfluxQL node page is to use one of those functions and see what functions I have at my disposal (since I usually just assume that almost anything can use `alert`, `deadman`, etc). It's definitely not a critical component, but I think it would be helpful to those getting started who aren't familiar with all of the InfluxDB functions available.

That being said, I think this is good to go as-is, so anything else is icing on the cake. :cake: We can always come back to it at a later date.

nathanielc:
@rossmcdonald 
> That being said, I think this is good to go as-is, so anything else is icing on the cake. :cake: We can always come back to it at a later date.

Let's do that.

rossmcdonald:
Sounds good. Merge when ready.

PR #311, opened 2016-03-14
Fix broken header linking

gunnaraasen:
cc @rossmcdonald @nathanielc 

This change fixes the broken header links. The CSS classes changed during the last redesign when search was added. This makes sure the javascript snippet that adds the header links refers to the correct classes.

@alexpaxton this change breaks the styling of h3 elements.


alexpaxton:
@gunnaraasen I'll jump in an fix styling on the `<h3>` links

Issue #315, opened 2016-03-15
Update Telegraf Documentation


rkuchan:
###### URL for relevant page?
https://docs.influxdata.com/telegraf/v0.10/

Current list includes:
* Support for retention policies (didn't mention)
* [x] Check plugins

PR #321, opened 2016-03-16
Telegraf 0.11 docs

rkuchan:
* Adds 0.11 docs for Telegraf
* Removes 0.10-specific references
* Updates the list of available plugins
* Makes 0.11 the default doc version for Telegraf

Fixes: https://github.com/influxdata/docs.influxdata.com/issues/315

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -433,85 +433,85 @@ SectionPagesMenu = "products"
         parent = "services"
 
 # Telegraf v0.10 Sub-Menu - Output Plugins
-[[menu.telegraf_011]]
+[[menu.telegraf_010]]
   name = "InfluxDB"
   identifier = "influxdb_output"
   weight = 0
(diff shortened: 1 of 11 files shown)


kostasb:
Telegraf 0.11 automatically starts on Ubuntu as part of the installation package, with this pull request:
https://github.com/influxdata/telegraf/pull/747

Cameron actually merged this commit to close #747 instead of merging the original PR:
https://github.com/influxdata/telegraf/commit/e4e7d7fbfc9dd244146d1ded02543f0775ac6988

I've tested and confirmed that the service does start automatically on Ubuntu and it doesn't start automatically on RPM-based (eg Centos), which is the desired behavior.

Maybe we should add a note for it under the "Start the Telegraf Server" section.

rkuchan:
@kostasb, thanks!

rkuchan:
Merging after in-person OK from @gunnaraasen.

Issue #274, opened 2016-02-25
Cluster updates


rkuchan:
[subject to change]

Relevant pages: https://docs.influxdata.com/influxdb/v0.10/clustering/

* [x] `INFLUXD_OPTS` in `/etc/default/influxdb` must reference every node including itself (unsure about non-Raft nodes)
* [x] Updated and improved `id` results from `SHOW SERVERS`
* [x] Configuration changes
* [x] Include information about open source clustering
* [x] Check https://github.com/influxdata/docs.influxdata.com/issues/325 (Issue closed)
 
Questions: Is this version of clustering still considered experimental?

Issue #155, opened 2016-01-15
Query Engine


benbjohnson:
The new query engine should be largely backwards compatible but I'm adding this issue to track differences.

https://github.com/influxdata/influxdb/pull/5196

benbjohnson:
`SELECT DISTINCT(value)` previously returned an array of values.  It will now return a separate row for each value.

rkuchan:
Queries with `LIMIT` and `SLIMIT` work - must be in that order.

PR #328, opened 2016-03-18
0.11 WIP

rkuchan:
[Subject to change]

* Cluster updates include (Fixes: https://github.com/influxdata/docs.influxdata.com/issues/274):
    * `INFLUXD_OPTS` in /etc/default/influxdb must reference every node including itself
    * Improved `id` numbers in `SHOW SERVERS`
    * A note about this being the last version with open source clustering
* Query language updates include (Fixes: https://github.com/influxdata/docs.influxdata.com/issues/155):
    * [`data_exploration.md`] Queries with `LIMIT` and `SLIMIT` work - must be in that order. Actually this looks like it worked in 0.10, we just didn't know about it.
    * [`functions.md`] `distinct()` now returns separate row for each value
    * [`functions.md`] `top()`, `bottom()` now return relevant timestamps with `GROUP BY <tag>` queries (instead of just the timestamp for the start of the `WHERE` time clause)
    * [`functions.md]  `first()` and `last()` no longer return the relevant timestamp.
    * [`schema_exploration.md`] `SHOW SERIES` output changes
    * [`schema_exploration.md`] `SHOW TAG VALUES` output changes (https://github.com/influxdata/influxdb/pull/5853)
* New page on the differences between 0.10 and 0.11 (`010_vs_011.md`), which covers:
    * Breaking API changes
    * New query engine
    * Storage engine requirements
    * Clustering updates
    * [Also: Removes the old differences pages (`08_vs_010.md` and `09_vs_010.md`) and creates a single page that links to them: `011_vs_previous.md`)
* Update the configuration file documentation
    * Add new cluster configuration settings in [meta] and [data]
    * Update directories ( `/opt/`s to `/lib/`s)
    * Remove references to non-TSM settings
    * Remove the [subscriber] section
    * Remove the duplicate [hinted-handoff] section
* Update the upgrade page (https://github.com/influxdata/docs.influxdata.com/issues/329)
    * Remove the CQ update section
    * Add a section on breaking API changes
    * Link back to 0.10 docs for the conversion tool (the tool doesn't work with 0.11)
* Final review
    * Fix several broken links
    * Remove Getting unexpected results with GROUP BY time() from the FEI page - I couldn't seem to repro with the new query engine
    * Add a note about the deprecated `IF EXISTS` syntax on `DROP DATABASE`
    * Link the to-come functions to https://github.com/influxdata/influxdb/issues/5930 instead of the individual issues


rkuchan:
@gunnaraasen, @beckettsean:

I'm still uncomfortable with the `Upgrading from previous versions` page as I haven't tested it out but other than that I've made every edit I could find for the 0.11 update.

I'm also assuming backups work the same in 0.11 as they did in 0.10.

PR #334, opened 2016-03-22
improve docs formatting

favadi:


kostasb:
This PR modifies the available time durations as a list on v0.10

duration_format

I am not sure about this formatting, we use the same for the "Note" just above it. I think the existing list look is easier to grasp.

rkuchan:
Thanks @favadi for your contribution! We're going to stick with the initial formatting for now but we appreciate your input!

PR #336, opened 2016-03-22
Clarity and consistency in the glossary of a "point"

zgohr:


beckettsean:
The tag set implies the series and vice versa, so these additions are redundant.

beckettsean:
I understand it might increase short-term comprehension to include the tag set, but that obscures the fact that the tag set and the series are one and the same (within a given measurement).

zgohr:
Ah. Well that's confusing. Thanks for the clarification. Might want to take a look [writing duplicate points](https://docs.influxdata.com/influxdb/v0.10/troubleshooting/frequently_encountered_issues/#writing-duplicate-points) in the FAQ, as it describes it using measurement/tag set and could potentially then be written as series instead. Or maybe a point isn't a duplicate if its same series, tag set, and different retention policy?

beckettsean:
Different retention policy does mean different point. In essence, the tag sets are scoped to a retention policy.

PR #338, opened 2016-03-23
Fix warning links from older to latest version for all products

kostasb:
Warning links for old versions are currently broken for all products.
This branch redirects the warning link for all v0.10 products to the same url path for latest / v0.11.
For versions older than 0.10 it just redirects to the root path of latest / v0.11 - with the exception of Kapacitor's webpage for 0.2 which also redirects to the full paths of latest / v0.11.

@rkuchan @gunnaraasen  staging for this branch:
http://docs.staging.influxdata.com:31006/


rkuchan:
+1!

Issue #343, opened 2016-03-23
Clarify the default interval for derivative on aggregates


jsternberg:
###### URL for relevant page?
https://docs.influxdata.com/influxdb/v0.11/query_language/functions/#derivative

###### What products and version are you using?
InfluxDB 0.11

###### Where did you look before opening the issue?
influxdata/influxdb#4819

The default interval for derivative when an aggregate is used is equal to the group by interval. So this query here:

```
SELECT derivative(mean(value)) FROM cpu WHERE time >= now() - 10m GROUP BY time(1m)
```

This is the same as:

```
SELECT derivative(mean(value), 1m) FROM cpu WHERE time >= now() - 10m GROUP BY time(1m)
```

beckettsean:
@jsternberg do you know when this behavior changed? It certainly used to default to 1s regardless of the GROUP BY interval. This is more sane, to be sure, but I'm curious why the change went through without anyone noticing.

jsternberg:
It might have been part of the query engine refactor since I'm not familiar with the old query engine. That issue was against 0.9.5 though so it may be older than that. Unfortunately, I don't know the history behind the behavior since my knowledge is limited to the new query engine.

jsternberg:
Just to clarify, the default for derivatives over fields is still 1 second. This only affects derivatives over aggregates. So this:

```
SELECT derivative(value) FROM cpu
```

Is the same as this:

```
SELECT derivative(value, 1s) FROM cpu
```

PR #347, opened 2016-03-23
Correct DERIVATIVE() and GROUP BY doc

rkuchan:
Clarifies that `DERIVATIVE()`'s rate of change defaults to the `GROUP BY time()` interval when the query includes a `GROUP BY time()` clause.

Fixes https://github.com/influxdata/docs.influxdata.com/issues/343


jackzampolin:
:+1:

PR #350, opened 2016-03-24
Explain how to use quotes

rubycut:
closes influxdata/kapacitor#367


nathanielc:
@rubycut Perfect thanks. 
@rkuchan This is good to go. Once merged, I'll update v0.12 too.

Issue #356, opened 2016-03-27
Incorrect field names


setharnold:
The Schema Design guide has a typo in the Tagset 2 column near the end, "field = 2 and region = north" is given twice. The second should be region midwest.

Sorry for lack of URL and more exact positioning, this is filed via mobile. Thanks.

rkuchan:
Thanks for pointing this out, @setharnold!

PR #358, opened 2016-03-29
Fix schema page typo in 0.10 and 0.11

rkuchan:
Fixes https://github.com/influxdata/docs.influxdata.com/issues/356.


gunnaraasen:
:+1:

PR #360, opened 2016-03-29
Kapacitor v0.12 Docs

nathanielc:
The docs are updated for the new TICKscript syntax. All references to the `.mapReduce` code have been removed.

TODO:
- [x] Wait for final 0.12 version of Kapacitor to be cut so that docs are in sync with code.


nathanielc:
These are good to go!

content/kapacitor/v0.12/index.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Kapacitor Version 0.12 Documentation
+
+menu:
+  kapacitor:
+    name: v0.12
+    identifier: kapacitor_012
+    weight: 0
```
gunnaraasen: This should be set to a value lower than the 0.11 weight.

gunnaraasen:
It's hard to tell exactly which content has changed but looks good from what I could tell (once the version weight is fixed).

PR #376, opened 2016-04-06
Add relay readme to docs

rkuchan:
Copies https://github.com/influxdata/influxdb-relay/blob/master/README.md into `high_availability/relay.md`.


rkuchan:
@toddboom, I noticed that there was content on docs.influxdata.com's Relay page yesterday. I think it was removed by the nightly docs build so I've added the relay README here - was that what was on the docs site yesterday?

PR #377, opened 2016-04-07
List 'ms' as valid unit for time()

cheppsn:
Milliseconds `ms` are also a valid unit for time().


beckettsean:
👍

rkuchan:
Thank you, @cheppsn!

PR #379, opened 2016-04-07
Fix documentation of a StreamNode, groupBy is a property, not a chain…

jonseymour:
…ing method.

Describing groupBy as a chaining method is misleading since in the context
of a stream node, groupBy is actually interpreted as a property - attempts
to use it as a chaining method result in a script compilation error.

The documentation is amended to reflect the actual behaviour.

Signed-off-by: Jon Seymour <jon@wildducktheories.com>


jonseymour:
Closed in favour of https://github.com/influxdata/kapacitor/pull/446 which fixes the problem at source.

PR #380, opened 2016-04-08
Adding link to the Telegraf windows service docs

ricardclau:
Should I also PR the other version folders?


beckettsean:
@ricardclau thanks for this! I don't think we should encourage Windows users to go with anything but the latest, given the embryonic state of Windows support for Telegraf.

Issue #372, opened 2016-04-04
Add note that compression is better with seconds precision


gunnaraasen:
This is new behavior with the TSM engine. I think we already have this documented in other places. This should also be mentioned on the line protocol page.

###### URL for relevant page?
https://docs.influxdata.com/influxdb/v0.11/write_protocols/line/#timestamp

###### What products and version are you using?
InfluxDB 0.10+

###### Where did you look before opening the issue?
As suggested in https://github.com/influxdata/influxdb/issues/6041#issuecomment-205067182.

Issue #354, opened 2016-03-25
Clarify that quotes need to be used for an identifier that conflicts with a keyword


jsternberg:
###### URL for relevant page?
https://docs.influxdata.com/influxdb/v0.11/query_language/spec/#identifiers

###### What products and version are you using?
InfluxDB 0.11

###### Where did you look before opening the issue?
influxdata/influxdb#6067

If we can just clarify in either the keywords or the identifiers section that double quotes are needed when a keyword conflicts with an identifier, that would be perfect.

Thanks!

PR #382, opened 2016-04-08
Fixing a couple issues

rkuchan:
* `data_exploration.md` was missing `ms` as a possible time precision - pointed out in https://github.com/influxdata/docs.influxdata.com/pull/377. This adds it to another place in 0.12 and to previous versions: 0.9, 0.10, and 0.11.

* Adds a note about double quoting identifiers if they contain a keyword to `spec.md`. Fixes https://github.com/influxdata/docs.influxdata.com/issues/354

* Adds a recommendation to use the smallest precision possible when writing data for compression purposed to 0.10, 0.11, and 0.12. Fixes 
https://github.com/influxdata/docs.influxdata.com/issues/372

### DIFF
--- a/content/influxdb/v0.10/write_protocols/line.md
+++ b/content/influxdb/v0.10/write_protocols/line.md
@@ -102,7 +102,7 @@ If the timestamp is not provided the point will inherit the server's local times
 
 Some write APIs allow passing a lower precision.
 If the API supports a lower precision, the timestamp may also be an integer epoch in microseconds, milliseconds, seconds, minutes or hours.
-We recommend using the smallest precision possible as this can result in
+We recommend using the least precise precision possible as this can result in
 significant improvements in compression.
 
 ## Full Example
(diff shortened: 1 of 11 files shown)


content/influxdb/v0.10/write_protocols/line.md
```diff
@@ -102,6 +102,8 @@ If the timestamp is not provided the point will inherit the server's local times
 
 Some write APIs allow passing a lower precision.
 If the API supports a lower precision, the timestamp may also be an integer epoch in microseconds, milliseconds, seconds, minutes or hours.
+We recommend using the smallest precision possible as this can result in
```
gunnaraasen: > smallest precision possible

May want to use a better descriptor here than "smallest". Maybe something like "least precise"?

gunnaraasen:
One comment, otherwise +1!

rkuchan:
Thanks, @gunnaraasen!

beckettsean:
I know it's already merged, but rather than "smallest" precision, we could say the "coarsest precision".

rkuchan:
I changed the wording on Gunnar's recommendation from smallest to 'least precise'. Do you prefer coarsest?

beckettsean:
It's great as-is, thanks.

PR #390, opened 2016-04-13
Add note to upgrading.md

rkuchan:
Adds a note about needing to change the `meta`, `data`, and `wal` directories in the newly-generated configuration file if the user is not using OS X.

By default, the directories are set to `/root/.influxdb` and InfluxDB will not start on their systems.


rkuchan:
@gunnaraasen, @jackzampolin

gunnaraasen:
+1

jackzampolin:
Thanks @rkuchan!

beckettsean:
@rkuchan I'm not sure I understand why this is needed. Shouldn't our upgrade process be easiest for the most common install type (ubuntu)? Why do we have OS X-specific logic in the upgrade process?

rkuchan:
@beckettsean, I'm not sure why but the default directories in a newly-generated config file are the OS X directories.

beckettsean:
@rkuchan even when generating the file on a linux system? That seems crazy.

rkuchan:
@beckettsean, yes. Agreed. Both Jack and I ran into this problem and it was totally unexpected.

PR #396, opened 2016-04-23
Typo error schema_and_data_layout.md "are are"

bloublou2014:
remove a double "are are"


rkuchan:
Thank you, @bloublou2014!

Issue #402, opened 2016-04-26
Style issue on the retention policy management page


mjdesa:
###### URL for relevant page?
https://docs.influxdata.com/influxdb/v0.12/query_language/database_management/#retention-policy-management

![unspecified](https://cloud.githubusercontent.com/assets/1477539/14833901/5b558264-0bb6-11e6-959a-f53bb6a8dcdc.png)

The `m minutes` should be on a new line.

PR #403, opened 2016-04-26
Fix funky formatting

rkuchan:
Fixes the `DURATION` formatting in verisons 0.9-0.12.

Fixes https://github.com/influxdata/docs.influxdata.com/issues/402


rkuchan:
@mjdesa, better?

Thank you for finding that 😬

mjdesa:
👍

PR #410, opened 2016-04-28
Fix typo in getting started TICKscript example

### LINKED ISSUE #288: Kapacitor Documentation is incomplete
I'm attempting to get started using Kapacitor, but having issue after issue. 

First, it appears that the 'real world' stream example just doesn't work
`invalid task: error calling func "percentile" on obj *kapacitor.influxqlMapReducers: reflect: Call using int64 as type float64`

From there, I attempted to find the documentation for percentile, and found nothing. 

Further, the sigma function says it uses a 'running mean', but the concept of running mean never defined.

These issues, and the general state of the documentation, are making it very hard to get using Kapacitor.

nathanielc:
Fixes #288

### DIFF
--- a/content/kapacitor/v0.10/introduction/getting_started.md
+++ b/content/kapacitor/v0.10/introduction/getting_started.md
@@ -354,7 +354,7 @@ stream
         .period(1m)
         .every(1m)
     // calculate the 95th percentile of the used cpu.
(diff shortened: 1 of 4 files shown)


nathanielc:
@beckettsean Simple typo fix in Kapacitor getting started docs.

beckettsean:
👍  merge away!

PR #423, opened 2016-05-04
Add note about permissions to upgrading.md

rkuchan:
@kostasb pointed this out:

The command that imports the meta directory in 0.12:

```
influxd restore -metadir=<path_to_012_meta_directory> <path_to_metastore_backup>
```

if run by the root user, will leave the `/var/lib/influxdb/meta/meta.db` file owned by root without write permission from anyone else.  I suspect that the InfluxDB service may not be able to write in the meta database if it starts with such permissions, but I’ve not seen or heard any issues so far. We should add a `chown influxdb:influxdb /var/lib/influxdb/meta/*` command.


beckettsean:
👍

PR #427, opened 2016-05-05
adding 0.11 dl links

### LINKED ISSUE #405: 0.12 upgrading should give 0.11 binary URL
###### URL for relevant page?
http://docs.influxdata.com/influxdb/v0.12/administration/upgrading/
upgrading from InfluxDB 0.9 and need to export/import the metastore via `backup`, which doesn't exist in 0.9. I can't use 0.12 since the metastore has changed. I can't find out how to download 0.11 in order to have the backup.

Should add at least 
https://dl.influxdata.com/influxdb/releases/influxdb_0.11.1-1_amd64.deb
and 
https://dl.influxdata.com/influxdb/releases/influxdb-0.11.1-1.x86_64.rpm
plus whatever the raw binary is for those not on apt or yum

beckettsean:
fixes #405

### DIFF
--- a/content/influxdb/v0.12/administration/upgrading.md
+++ b/content/influxdb/v0.12/administration/upgrading.md
@@ -32,7 +32,7 @@ file.
 
 To start out, you must be working with version 0.10 or 0.11 (don't upgrade the
 `influxd` binary yet!).
-If you've already upgraded the binary, reinstall 0.10 or 0.11 (see the end of the page for links); InfluxDB 0.12
+If you've already upgraded the binary, reinstall 0.10 or 0.11 (see the [end of the page](/influxdb/v0.12/administration/upgrading/#urls-for-influxdb-0-11) for links); InfluxDB 0.12
 will yield an error
 (`run: create server: detected /var/lib/influxdb/meta/raft.db. [...]`) if you
 attempt to start the process without completing the steps below.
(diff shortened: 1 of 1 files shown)


beckettsean:
@rkuchan @gunnaraasen @rossmcdonald 

The one remaining question is what advice to give Homebrew users. I don't know how they would fetch 0.11.1 from Homebrew if they already upgraded to 0.12. Thoughts?

rossmcdonald:
I'm not sure if you can easily downgrade using Homebrew alone. As a worst case scenario, they can always download the binaries manually:

```
# Stop 0.12 server

# Download and start 0.11 server
wget https://dl.influxdata.com/influxdb/releases/influxdb-0.11.1-1_darwin_amd64.tar.gz
tar -xvf influxdb-0.11.1-1_darwin_amd64.tar.gz
./influxdb-0.11.1-1/usr/bin/influxd

# In a new window, run the backup
./influxdb-0.11.1-1/usr/bin/influxd backup /tmp/backup
```

beckettsean:
updated with better formatting and @rossmcdonald's suggestions.

gunnaraasen:
LGTM I've been wanting to set up our own Homebrew tap with older versions which would make rolling back to an earlier version much easier.

PR #446, opened 2016-05-14
Update Line Protocol example code documentation

Starefossen:
This PR expands the documentation explaining the Full Example for the Line Protocol.

### DIFF
--- a/content/influxdb/v0.13/write_protocols/line.md
+++ b/content/influxdb/v0.13/write_protocols/line.md
@@ -106,15 +106,21 @@ We recommend using the least precise precision possible as this can result in
 significant improvements in compression.
 
 ## Full Example
+
 A full example is shown below.
+
 ```
 cpu,host=server01,region=uswest value=1 1434055562000000000
 cpu,host=server02,region=uswest value=3 1434055562000010000
 temperature,machine=unit42,type=assembly internal=32,external=100 1434055562000000035
 temperature,machine=unit143,type=assembly internal=22,external=130 1434055562005000035
 ```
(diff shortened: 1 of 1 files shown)


content/influxdb/v0.13/write_protocols/line.md
```diff
@@ -106,15 +106,21 @@ We recommend using the least precise precision possible as this can result in
...
 cpu,host=server02,region=uswest value=3 1434055562000010000
 temperature,machine=unit42,type=assembly internal=32,external=100 1434055562000000035
 temperature,machine=unit143,type=assembly internal=22,external=130 1434055562005000035
 ```
-In this example the first line shows a `measurement` of "cpu", there are two tags "host" and "region", the `value` is 1.0, and the `timestamp` is 1434055562000000000.
+
+In this example the first line shows a `measurement` of "cpu", there are two `tags` "host" and "region", the `value` is 1.0, and the `timestamp` is 1434055562000000000.
 Following this is a second line, also a point in the `measurement` "cpu" but belonging to a different "host".
+
+The two last lines of the example shows two `measurements` of "temperature", there are two `tags` "machine" and "type", with `values` for "internal" and "external" temperature for both measurements.
```
rkuchan: I would slightly reword this last bit to something like this:

The two last lines of the example show two points in the `measurement` "temperature". There are two `tags` "machine" and "type" with `fields` for "internal" and "external" temperatures.

rkuchan:
Thank you, @Starefossen! I made a small edit but once that's taken care of we can get this merged!

Starefossen:
Fixed 😄

rkuchan:
Awesome. Thanks again, @Starefossen!

PR #451, opened 2016-05-18
Make the downloads page open in new window

rkuchan:
Fixes https://github.com/influxdata/docs.influxdata.com/issues/431

Props to @wfro for helping me fix this 😇


gunnaraasen:
+1

PR #452, opened 2016-05-18
Add continuous queries example to Kapacitor

nathanielc:
This was already a blog post but I thought it would be good to add to the docs examples too.


rossmcdonald:
+1, great addition to the docs. I also like the stream vs. batch comparison.

Issue #453, opened 2016-05-18
Document SOFFSET


rkuchan:
In both `spec.md` and `data_exploration`.

Issue #456, opened 2016-05-18
GROUP BY time() offset interval allows -


rkuchan:
data_exploration, 0.13

```
> SELECT COUNT(water_level) FROM h2o_feet WHERE time >= '2015-08-19T00:00:00Z' AND time <= '2015-08-27T17:00:00Z' AND location='coyote_creek' GROUP BY time(3d,-15m)
name: h2o_feet
--------------
time			count
2015-08-17T23:45:00Z	478
2015-08-20T23:45:00Z	720
2015-08-23T23:45:00Z	720
2015-08-26T23:45:00Z	173
```

PR #457, opened 2016-05-18
Update Data Exploration

rkuchan:
This adds a description of `SOFFSET` and an example of a negative offset interval in a `GROUP BY time()` query to `data_exploration.md`.

It also fixes the syntax in `spec.md` for the SHOW TAG VALUES query. It adds the double quotes to `data_exploration.md` as well.

Fixes https://github.com/influxdata/docs.influxdata.com/issues/453, https://github.com/influxdata/docs.influxdata.com/issues/456, and https://github.com/influxdata/docs.influxdata.com/issues/460.


beckettsean:
LGTM!

Issue #454, opened 2016-05-18
Clarify HTTP batch size recommendations


gunnaraasen:
We typically recommend a batch size of 5000 points per HTTP request. Our documentation seems to suggest that a 5000 point batch is the upper limit instead of the recommended size. We should state that we recommend a 5000 to 10000 point batch size and that smaller batches (and more HTTP requests) will lead to sub-optimal performance.

###### URL for relevant page?
https://docs.influxdata.com/influxdb/v0.13/guides/writing_data/
https://docs.influxdata.com/influxdb/v0.13/write_protocols/write_syntax/#http

###### What products and version are you using?
InfluxDB 0.13

###### Where did you look before opening the issue?
https://groups.google.com/forum/#!searchin/influxdb/batch$20size%7Csort:date

Issue #442, opened 2016-05-12
clean up write syntax page


beckettsean:
###### URL for relevant page?
https://docs.influxdata.com/influxdb/v0.13/write_protocols/write_syntax/
###### What products and version are you using?
0.13

Make the layout better, probably tabular.

Issue #441, opened 2016-05-12
clarify GET vs POST for /query endpoint


beckettsean:
###### What products and version are you using?
0.13+

PR #462, opened 2016-05-19
Edit API doc

rkuchan:
This takes a stab at an API reference doc. It adds detail to the API doc in the Concepts section. It covers each endpoint (`/ping`,`query`, and `/write`) and, where applicable, goes into detail about the required and optional query string parameters, status codes and responses, and example requests. 

The sections on status codes and query string parameters need special attention - I included things I was able to test but I have a feeling I am missing some.

I would like this to live somewhere else on the docs but that discussion can occur offline as it involves a significant reorganization of the sidebar.

This fixes:
* https://github.com/influxdata/docs.influxdata.com/issues/454 (Clarify batch size recommendations)
* https://github.com/influxdata/docs.influxdata.com/issues/441 (Clarify `GET`/`POST` for the `/query` endpoint)
* https://github.com/influxdata/docs.influxdata.com/issues/442 (Clean up write syntax) - I don't actually touch the write syntax page in this PR but a sidebar reorganization would make this the main API doc.

### DIFF
--- a/content/influxdb/v0.13/concepts/api.md
+++ b/content/influxdb/v0.13/concepts/api.md
@@ -1,12 +1,10 @@
 ---
-title: API Endpoints & Ports
+title: API Reference
 menu:
   influxdb_013:
     weight: 80
     parent: concepts
 ---
-
-## API reference
 
 The InfluxDB API provides a simple way interact with the database.
 It uses HTTP response codes, HTTP authentication, and responses are returned in
@@ -102,9 +100,10 @@ Those `SELECT` queries require a `POST` request.
 --data-urlencode "q=<InfluxQL query>"
 ```
 
-Use curl's `--data-urlencode` encoding method for all queries.
-Queries must be in [InfluxQL](influxdb/v0.13/query_language/), InfluxDB's query
-language.
+All queries must be URL encoded and follow
+[InfluxQL](/influxdb/v0.13/query_language/) syntax.
+Our example shows the `--data-urlencode` parameter from `curl`, which we will
+use in all examples on this page.
 
 Delimit multiple queries with a semicolon.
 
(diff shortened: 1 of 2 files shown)


content/influxdb/v0.13/concepts/api.md
```diff
@@ -26,32 +51,341 @@ X-Influxdb-Version: 0.13.x
...
+| rp=\<retention_policy_name> | Optional | Sets the target [retention policy](/influxdb/v0.13/concepts/glossary/#retention-policy-rp) for the query. InfluxDB queries the `DEFAULT` retention policy if you do not specify a retention policy.  |
+| u=\<username> | Optional if you haven't [enabled authentication](/influxdb/v0.13/administration/authentication_and_authorization/#set-up-authentication). Required if you've enabled authentication. | Sets the username for authentication if you've enabled authentication. The user must have read access to the database. Use with the query string parameter `p`. |
+
+#### Request Body
+
+```
+--data-urlencode "q=<InfluxQL query>"
+```
+
+Use curl's `--data-urlencode` encoding method for all queries.
```
beckettsean: URL encoding isn't something specific to `curl`. Let's reword to something like:

All queries myst be URL encoded and in [InfluxQL](influxdb/v0.13/query_language/), InfluxDB's query language. 
Our example show's the `--data-urlencode` parameter from `curl`, which we will use in all examples on this page.

content/influxdb/v0.13/concepts/api.md
```diff
@@ -26,32 +51,341 @@ X-Influxdb-Version: 0.13.x
...
+| u=\<username> | Optional if you haven't [enabled authentication](/influxdb/v0.13/administration/authentication_and_authorization/#set-up-authentication). Required if you've enabled authentication. | Sets the username for authentication if you've enabled authentication. The user must have write access to the database. Use with the query string parameter `p`. |
+
+
+#### Request Body
+
+```
+--data-binary '<Data in Line Protocol format>'
+```
+
+Use curl's `--data-binary` encoding method for all writes.
```
beckettsean: Similar to above, binary encoding isn't specific to `curl`. 

All line protocol writes must be binary encoded and in the [Line Protocol format](https://docs.influxdata.com/influxdb/v0.13/write_protocols/line/).
Our example shows the `--data-binary` parameter from curl, which we will use in all examples on this page.

content/influxdb/v0.13/concepts/api.md
```diff
@@ -26,32 +51,341 @@ X-Influxdb-Version: 0.13.x
...
-### Admin interface port
+##### Summary Table
+<br>
 
-The admin interface for InfluxDB runs on port `8083` and exposes web UI for the server.
+| HTTP status code | Description    |
+| :--------------- | :------------- |
+| 204 No Content   | Success!      |
+| 400 Bad Request  | Unacceptable request. Can occur with a Line Protocol syntax error or if a user attempts to write values to a field that previously accepted a different value type. The returned JSON offers further information. |
+| 404 Not Found    | Unacceptable request. Can occur a user attempts to write to a database that does not exist. The returned JSON offers further information. |
```
beckettsean: Can occur **if** a user attempts...

beckettsean:
great doc!

rkuchan:
Thanks, @beckettsean!

PR #465, opened 2016-05-22
Proper invert match operator in Query Language Reference

ecneladis:
Based on current source code: https://github.com/influxdata/influxdb/blob/master/influxql/scanner_test.go#L59


rkuchan:
Thank you, @ecneladis!

PR #470, opened 2016-05-25
Fix query example to filter when tag has no value

titilambert:


titilambert:
I'm not sure if it's a Influxdb issue or doc issue ...

beckettsean:
@titilambert I don't understand what you are trying to fix. The syntax in the current docs is correct. Your syntax would change the regex from matching any string to matching any single character.

Why did you think the examples were in correct?

rkuchan:
@beckettsean, I tested this out and found that including the `*` doesn't return what I'd expect:

```
> insert mymeas,tag1=1 source=1
> insert mymeas,tag1=1 source=1
> insert mymeas,tag1=1 source=1
> insert mymeas,tag2=1 source=1

> select * from mymeas where tag1 !~ /.*/
> select * from mymeas where tag1 !~ /./
name: mymeas
------------
time				            source	tag1	tag2
2016-05-25T15:32:46.014225708Z 	1		        1

> select * from mymeas where tag1 =~ /.*/
name: mymeas
------------
time				            source	tag1	tag2
2016-05-25T15:32:38.719795629Z	1	    1
2016-05-25T15:32:40.694698006Z	1	    1
2016-05-25T15:32:42.306488816Z	1	    1
2016-05-25T15:32:46.014225708Z	1		        1

> select * from mymeas where tag1 =~ /./
name: mymeas
------------
time				            source	tag1	tag2
2016-05-25T15:32:38.719795629Z	1	    1
2016-05-25T15:32:40.694698006Z	1	    1
2016-05-25T15:32:42.306488816Z	1	    1
```

beckettsean:
Wow, that looks like a regression in the query engine. Maybe Go's regex parser changed?

@jwilder @benbjohnson @jsternberg thoughts on why the behavior would have changed?

Some experimentation suggests that `.*` now matches the empty string, which would explain why the negation no longer works:

```
> select * from regex
name: regex
-----------
time			t1	val
1464196328279739372	aaa	1
1464196333328521644	bbb	2
1464196337464012158		3

> select * from regex where t1 =~ /.*/
name: regex
-----------
time			t1	val
1464196328279739372	aaa	1
1464196333328521644	bbb	2
1464196337464012158		3

> select * from regex where t1 =~ /./
name: regex
-----------
time			t1	val
1464196328279739372	aaa	1
1464196333328521644	bbb	2

> select * from regex where t1 !~ /./
name: regex
-----------
time			t1	val
1464196337464012158		3

```

beckettsean:
Verified this is a change in behavior from 0.12:

```
# influx
Visit https://enterprise.influxdata.com to register for updates, InfluxDB server management, and monitoring.
Connected to http://localhost:8086 version 0.12.2
InfluxDB shell 0.12.2
> create database mydb
> use mydb
Using database mydb
> insert regex,t1=aaa val=1
> insert regex,t1=bbb val=2
> insert regex val=3
> select * from regex where t1 =~ /.*/
name: regex
-----------
time			t1	val
1464196545410584111	aaa	1
1464196550225990061	bbb	2

> select * from regex where t1 =~ /./
name: regex
-----------
time			t1	val
1464196545410584111	aaa	1
1464196550225990061	bbb	2

> select * from regex where t1 !~ /./
name: regex
-----------
time			t1	val
1464196555286367788		3

> select * from regex where t1 !~ /.*/
name: regex
-----------
time			t1	val
1464196555286367788		3

```

beckettsean:
I'm less concerned about updating the docs with the "new" syntax than I am with understanding what changed. I'll open an issue on the database so we can find out why this changed and decide whether it was intentional, unintentional but desired, or something we should revert.

jsternberg:
This was an intentional change to my knowledge. I think it was #3773 and it seems to have been introduced as a bugfix for 0.13. A non-existant tag is supposed to be the same as an empty tag (as there's no difference between an empty tag and a non-existant tag in the system to my knowledge).

beckettsean:
@jsternberg I agree that a non-existent tag and an empty tag should be the same thing. The question is why does `/.*/` now match an empty tag, when before it did not?

https://github.com/influxdata/influxdb/issues/6726

jsternberg:
@beckettsean it's because of influxdata/influxdb#3773.

jsternberg:
Sorry, that's the issue. It was fixed in influxdata/influxdb#6283.

titilambert:
So this is a Influxdb issue, not bad documentation :)

beckettsean:
@jsternberg so the new intent is that:

`foo =~ /./` should match any non-empty value
`foo !~ /./` should match any empty/non-existent value
`foo =~ /.*/` should match anything
`foo ~~ /.*/` should not match anything

So now `/.*/` is basically a useless regex? 

From looking at the code, it appears that any regex more complex than these was unaffected, so `/foo.$/` should match `fooo` but not `foo123` whereas `/foo.*$/` should match both. Is that still correct?

jsternberg:
You are correct. Except the small typo for the fourth example where you said `~~` instead of `!~`.

And yes, `/.*/` is basically a useless regex.

beckettsean:
Gotcha. I fixed my comment typo, and we will accept this PR to make the docs correct with the current parsing implementation.

beckettsean:
Thanks @titilambert!

PR #492, opened 2016-06-06
Avoid using a keyword as a measurement name in an example.

jonseymour:
If the reader attempts to use a measurement with a name of "measurement"
in a query then they will likely omit the required double-quoting of
measurement where it is used as a measurement name and then receive a
confusing report of syntax error as a result.

With this change, the measurement name used in the example will not
require quoting if subsequently used.

See https://github.com/influxdata/influxdb/issues/6785

Signed-off-by: Jon Seymour <jon@wildducktheories.com>

### DIFF
--- a/content/influxdb/v0.13/write_protocols/write_syntax.md
+++ b/content/influxdb/v0.13/write_protocols/write_syntax.md
@@ -17,7 +17,7 @@ The syntax for the line protocol is
 For example:
 
 ```bash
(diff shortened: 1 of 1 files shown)


beckettsean:
@jonseymour thanks for the PR! I believe that section is undergoing a substantial re-write currently. 

@rkuchan would the updates here survive the re-write of `/write_protocols/write_syntax.md`?

rkuchan:
@jonseymour, thanks for putting that together! Merged into the 0.13 docs.

PR #445, opened 2016-05-14
Rk cq guides

rkuchan:
* Creates a new CQ use cases guide
* Adds Downsampling and Data Retention to that guide, and makes it a lot shorter, this fixes https://github.com/influxdata/docs.influxdata.com/issues/447
* Adds a section about using CQs in places of HAVING or nested functions; this fixes https://github.com/influxdata/docs.influxdata.com/issues/351

PR #499, opened 2016-06-08
Edit and condense the downsampling guide

rkuchan:
This was originally part of https://github.com/influxdata/docs.influxdata.com/pull/445, but we decided to keep the downsampling guide as a separate guide.

The PR makes all the edits that Sean suggested in https://github.com/influxdata/docs.influxdata.com/pull/445.

### DIFF
--- a/content/influxdb/v1.0/guides/downsampling_and_retention.md
+++ b/content/influxdb/v1.0/guides/downsampling_and_retention.md
@@ -60,16 +60,17 @@ Assume that, in the long run, we're only interested in the average number of ord
 and by website at 30 minute intervals.
 In the next steps, we use RPs and CQs to:
 
- * Automatically delete the raw, ten-second resolution data that are older than two hours
  * Automatically aggregate the ten-second resolution data to 30-minute resolution data
+ * Automatically delete the raw, ten-second resolution data that are older than two hours
  * Automatically delete the 30-minute resolution data that are older than 52 weeks
 
 ### Database Preparation
 We perform the following steps before writing the data to the database
 `food_data`.
-We do this **before** inserting any data because InfluxDB only performs CQs on
-new data, that is, data with timestamps that occur after the initial execution
-of the CQ.
+We do this **before** inserting any data because CQs only run against recent
+data; that is, data with timestamps that are no older than `now()` minus
+the `FOR` clause of the CQ, or `now()` minus the `GROUP BY time()` interval if
+the CQ has no `FOR` clause.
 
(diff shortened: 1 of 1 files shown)


content/influxdb/v1.0/guides/downsampling_and_retention.md
```diff
@@ -7,143 +7,221 @@ menu:
...
+### Goal
+Assume that, in the long run, we're only interested in the average number of orders by phone
+and by website at 30 minute intervals.
+In the next steps, we use RPs and CQs to:
 
-### Purpose
-In general, InfluxDB wasn't built to process deletes.
-One of the fundamental assumptions in its architecture is that deletes are infrequent and need not be highly performant.
-However, InfluxDB recognizes the necessity of purging data that have outlived their usefulness - that is the purpose of RPs.
+ * Automatically delete the raw, ten-second resolution data that are older than two hours
```
beckettsean: I would switch the order to be:

* Automatically aggregate the ten-second resolution data to 30-minute resolution data
* Automatically delete the raw, ten-second resolution data that are older than two hours
* Automatically delete the 30-minute resolution data that are older than 52 weeks

content/influxdb/v1.0/guides/downsampling_and_retention.md
```diff
@@ -7,143 +7,221 @@ menu:
...
+ * Automatically aggregate the ten-second resolution data to 30-minute resolution data
+ * Automatically delete the 30-minute resolution data that are older than 52 weeks
 
-### Working with RPs
-When you create a database, InfluxDB automatically creates an RP called `autogen` with an infinite duration and a replication factor set to one.
-`autogen` also serves as the `DEFAULT` RP; if you do not supply an explicit RP when you write a point to the database, the data are subject to the `DEFAULT` RP.
+### Database Preparation
+We perform the following steps before writing the data to the database
+`food_data`.
+We do this **before** inserting any data because InfluxDB only performs CQs on
```
beckettsean: mostly true, but not quite. If I create a CQ with `EVERY 10m FOR 5d` I will grab data up to 5 days old, even if the CQ is created today.

"We do this before inserting any data because CQs only run against recent data; that is, data with timestamps no older than `now()` minus the `FOR` clause of the CQ, or `now()` minus the `GROUP BY time()` interval if the CQ has no `FOR` clause."

beckettsean:
minor notes, otherwise good to go!

rkuchan:
Thanks, @beckettsean!

PR #502, opened 2016-06-14
Add systemd commands to InfluxDB installation

### LINKED ISSUE #494: add installing doc on ubuntu 16 (systemd)
###### URL for relevant page?
https://docs.influxdata.com/influxdb/v0.13/introduction/installation/

in ubuntu 16 instead of doing this to start the service:
```
sudo apt-get update && sudo apt-get install influxdb
sudo service influxdb start
```

you need to use systemctl
to start the service:
`systemctl start influxdb`

check is status:
`systemctl status influxdb`

make it run on boot
`systemctl enable influxdb`

you can see for example how grafana docs handle it:
http://docs.grafana.org/installation/debian/#start-the-server-via-systemd

ubuntu 14 lts end of life is in August 2016 and I guess that a lot will start to use ubuntu 16 lts lts which uses systemctl

rossmcdonald:
Update InfluxDB installation documentation to include systemd commands.

Fixes #494

### DIFF
--- a/content/influxdb/v0.13/introduction/installation.md
+++ b/content/influxdb/v0.13/introduction/installation.md
@@ -32,9 +32,12 @@ you may want to check out our
 
 ### Ubuntu & Debian
 
-For instructions on how to install the Debian package from a file, please see the [downloads page](https://influxdata.com/downloads/).
+For instructions on how to install the Debian package from a file,
+please see the
+[downloads page](https://influxdata.com/downloads/). Debian and Ubuntu
+users can install the latest stable version of InfluxDB using the
+`apt-get` package manager.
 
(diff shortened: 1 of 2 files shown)


content/influxdb/v0.13/introduction/installation.md
```diff
@@ -59,6 +62,13 @@ sudo apt-get update && sudo apt-get install influxdb
 sudo service influxdb start
 ```
 
+Or if your operating system is using systemd:
```
gunnaraasen: Might want to list the OS versions which use systemd by default.

> Or if your operating system is using systemd (Ubuntu 15.04+, Debian 8+):

rossmcdonald: Good idea. Updated.

PR #515, opened 2016-06-28
Update docs for octal number in TICKscript

nathanielc:
As per https://github.com/influxdata/kapacitor/pull/684 TICKscript support octal number literal. This updates the docs accordingly.


rossmcdonald:
👍

Issue #516, opened 2016-06-29
Grammartical Error


Onuonga:
The error is specifically at: https://docs.influxdata.com/influxdb/v0.13/guides/writing_data/ 
This is based of the docs that are on influx db - Writing data (version 0.1.3 )

beckettsean:
@Onuonga that's a fairly large page. Can you specify what grammatical error you see and where?

beckettsean:
Data is commonly used in the plural sense, especially in documentation. Either is acceptable and correct.

PR #517, opened 2016-06-29
-- Fix Grammatical Issue --

Onuonga:
* You can not use those in this case. Switched it to that. The data is referred to as a whole as "that data".
  
Example: That data is corrupt and *not* those data is corrupt

fixes issue #516

### DIFF
--- a/content/influxdb/v0.13/guides/writing_data.md
+++ b/content/influxdb/v0.13/guides/writing_data.md
@@ -84,7 +84,7 @@ InfluxDB will still attempt to write the points after that time out but there wi
 ---
 InfluxDB is a schemaless database.
 You can add new measurements, tags, and fields at any time.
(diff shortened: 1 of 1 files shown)


beckettsean:
In the InfluxDB documentation, "data" is used in the plural sense.

From http://www.merriam-webster.com/dictionary/data (emphasis mine):

Usage Discussion of data
Data leads a life of its own quite independent of datum, of which it was originally the plural. It occurs in two constructions: as a plural noun (like earnings), **taking a plural verb and plural modifiers (as these, many, a few)** but not cardinal numbers, and serving as a referent for plural pronouns (as they, them); and as an abstract mass noun (like information), taking a singular verb and singular modifiers (as this, much, little), and being referred to by a singular pronoun (it). Both constructions are standard. **The plural construction is more common in print**, evidently because the house style of several publishers mandates it.

PR #530, opened 2016-07-14
Fix the chunking documentation

jsternberg:
The chunking documentation needs to mention the `chunked=true` parameter
is required and that chunking is done by either series or the maximum
size.

Also updated the documentation to explain what `max-row-limit` does and
when it applies.

Fixes influxdata/influxdb#6999.


beckettsean:
thanks, @jsternberg!

PR #534, opened 2016-07-19
kapacitor beta3 updates

nathanielc:


nathanielc:
@beckettsean @rossmcdonald How does this look?

beckettsean:
looks good as far as I understand it.

PR #554, opened 2016-07-28
Update version id to xxxx_1_0

rkuchan:
For future ease, this changes the internal version id for 1.0 from `<product>_1` to `<product>_1_0`.


rkuchan:
@adamperlin, here you go!

PR #571, opened 2016-08-06
use kleene star for repeated fields

anfedorov:


beckettsean:
Thanks for the suggestion, @anfedorov. We've been consistent in using ellipses throughout the documentation and would prefer to keep that convention for now.

@rkuchan

Issue #584, opened 2016-08-24
New mode function


jwilder:
New `mode` function was added for 1.0.

###### URL for relevant page?

https://github.com/influxdata/influxdb/pull/7199

###### What products and version are you using?

InfluxDB 1.0

###### Where did you look before opening the issue?

https://docs.influxdata.com/influxdb/v0.13/query_language/functions/

PR #585, opened 2016-08-24
Add mode() function to docs

agaurav:
Closes https://github.com/influxdata/docs.influxdata.com/issues/584


rkuchan:
@agaurav, this is great! I have two small requests and once you do those we'll get this merged:

* Could you replace the CLI responses with the following? Sometimes the spacing doesn't quite match up on our docs site and these should do it.  
    For the first example:
    ```bash
name: h2o_feet
--------------
time                   mode
1970-01-01T00:00:00Z   2.69
    ```
    For the second example:
    ```bash
name: h2o_feet
tags: location=coyote_creek
time                   median
----                   ------
2015-08-18T00:00:00Z   7.8245

name: h2o_feet
tags: location=santa_monica
time                   median
----                   ------
2015-08-18T00:00:00Z   2.0575
    ```
* Could you take a look at the merge conflicts?

Thanks again for putting this together!

agaurav:
@rkuchan done.

Issue #592, opened 2016-08-26
update FEQ for 1.0, remove bug that's no longer there


beckettsean:
https://github.com/influxdata/influxdb/issues/2703 has been fixed: https://github.com/influxdata/influxdb/issues/2703#issuecomment-242839037

###### URL for relevant page?
https://docs.influxdata.com/influxdb/v1.0/troubleshooting/frequently_encountered_issues/#querying-a-time-range-that-spans-epoch-0

Issue #583, opened 2016-08-24
remove support for negative timestamps


beckettsean:
Query engine refactor in 0.11 appears to have disabled negative timestamps in InfluxQL. They can still be written, but not queried. The WHERE clause won't accept negative timestamps, relative or absolute, and the default lower bound is epoch 0. 

###### URL for relevant page?
https://docs.influxdata.com/influxdb/v0.11/troubleshooting/frequently_encountered_issues/#querying-outside-the-min-max-time-range
https://docs.influxdata.com/influxdb/v0.11/troubleshooting/frequently_encountered_issues/#querying-a-time-range-that-spans-epoch-0
https://docs.influxdata.com/influxdb/v0.12/troubleshooting/frequently_encountered_issues/#querying-outside-the-min-max-time-range
https://docs.influxdata.com/influxdb/v0.12/troubleshooting/frequently_encountered_issues/#querying-a-time-range-that-spans-epoch-0
https://docs.influxdata.com/influxdb/v0.13/troubleshooting/frequently_encountered_issues/#querying-a-time-range-that-spans-epoch-0
https://docs.influxdata.com/influxdb/v1.0/troubleshooting/frequently_encountered_issues/#querying-a-time-range-that-spans-epoch-0
possibly others

Issue #552, opened 2016-07-27
need a page in InfluxDB docs about clustering


beckettsean:
It's our #1 search term, we need to send users somewhere helpful, like the new Enterprise docs, or for older versions, to Paul's blog post, perhaps.

pauldix:
I'd rather not bother to send them to the blog post. Let's just have a landing page in the docs for when they search on clustering. Something like this should be enough.

> Clustering is offered in our commercial Enterprise offering of InfluxDB. Go [here](https://portal.influxdata.com/) for more details.

beckettsean:
Updated copy:

> For high availability or scale out clustering we have a commercial offering called InfluxEnterprise. For details go here.

beckettsean:
Final text for the page:

============
Open-source InfluxDB does not support clustering. For high availability or horizontal scaling of InfluxDB, please investigate our commercial clustered offering, [InfluxEnterprise](https://portal.influxdata.com).

beckettsean:
@rkuchan can you add the same page to the 0.13 docs, too?

PR #593, opened 2016-08-26
A couple small fixes

rkuchan:
Fixes:

* https://github.com/influxdata/docs.influxdata.com/issues/552 (Adds cluster page to 0.13 docs)
* https://github.com/influxdata/docs.influxdata.com/issues/583 (Adds FEI about negative timestamps being write-able but not query-able to versions 0.11 through 0.13)
* https://github.com/influxdata/docs.influxdata.com/issues/592 (Removes FEI about not being able to query timestamps that span epoch 0)


beckettsean:
👍

PR #615, opened 2016-08-29
proposed changes to eval docs

beckettsean:


beckettsean:
@rossmcdonald @nathanielc

nathanielc:
@beckettsean These changes look good. But since they are part of the auto generated docs they need to be changed in the Kapacitor source in order to be permanent. I'll port these changes over to a Kapacitor PR and then submit a new docs PR after generating.

Thanks!

beckettsean:
Thanks, @nathanielc. Do we need to do the same for https://github.com/influxdata/docs.influxdata.com/pull/618? That doesn't seem auto-generated

nathanielc:
See https://github.com/influxdata/kapacitor/pull/846

nathanielc:
@beckettsean Nope, only docs under a `nodes` directory are auto generated.  Thanks for checking. I'll take a look at #618.

PR #624, opened 2016-08-30
Attempt at clarifying kapacitor's window node.

goller:
I was trying to clarifying window nodes for kapacitor... I definitely need input on this.


beckettsean:
A couple small changes, otherwise looks great!

PR #625, opened 2016-08-30
correct grammar

jademcgough:
I believe this content is duplicated in every kapacitor node doc, so we'd need to make sure that this propagates to those as well.

### DIFF
--- a/content/kapacitor/v1.0/nodes/union_node.md
+++ b/content/kapacitor/v1.0/nodes/union_node.md
@@ -137,7 +137,7 @@ Returns: [InfluxQLNode](/kapacitor/v1.0/nodes/influx_q_l_node/)
 
 ### Combine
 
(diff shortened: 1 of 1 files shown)


beckettsean:
@nathanielc more for the Kap source

beckettsean:
@nathanielc perhaps you can show @rkuchan and I where to change this so we can update the Kap docs and leave you out of the critical path?

nathanielc:
@rkuchan @beckettsean The docs just come from comments in the code. All the relevant code lives in this directory https://github.com/influxdata/kapacitor/tree/master/pipeline 

For example https://github.com/influxdata/kapacitor/blob/master/pipeline/node.go#L357

I would appreciate a PR against Kapacitor with these changes :)

jademcgough:
moving this to https://github.com/influxdata/kapacitor/pull/852

Issue #621, opened 2016-08-30
move clustering page to HA section


beckettsean:
###### URL for relevant page?
https://docs.influxdata.com/influxdb/v0.13/concepts/clusters/
https://docs.influxdata.com/influxdb/v1.0/concepts/clusters/

Let's move these alongside Relay in the High Availability section. New URLs would be:

https://docs.influxdata.com/influxdb/v0.13/high_availability/clusters/
https://docs.influxdata.com/influxdb/v1.0/high_availability/clusters/

Issue #623, opened 2016-08-30
Outdated FEI about the chunked parameter


rkuchan:
https://docs.influxdata.com/influxdb/v1.0/troubleshooting/frequently_encountered_issues/#getting-large-query-returns-in-batches-when-using-the-http-api

Need to update in 0.13 as well.

PR #627, opened 2016-08-30
Rk some fixes

rkuchan:
Moves the clustering page to the HA section and removes the FEI about the outdated `chunk_size` query string parameter.

Fixes: #621 and #623


beckettsean:
👍

PR #632, opened 2016-08-31
fix license required statement

kostasb:
@rkuchan 
A valid license is mandatory for influxd-meta and influxd to work, right?
The docs stated that the license is *not* required to start these processes, I am not sure if I get it right though.


rkuchan:
@kostasb - oh my, good find. Thank you for fixing that.

Issue #148, opened 2016-01-14
add series cardinality RAM issues to FEI page


beckettsean:
Describe that high series cardinality can lead to OOM, and discuss how to reduce series cardinality (`DROP MEASUREMENT`?)

PR #648, opened 2016-09-02
add OOM due to series cardinality to FEI

beckettsean:
fixes https://github.com/influxdata/docs.influxdata.com/issues/148


rkuchan:
👍

PR #661, opened 2016-09-09
Update location of default config file

dandv:


beckettsean:
@rossmcdonald

dandv:
Thanks Ross. I didn't see `/etc/default/influxdb` at https://docs.influxdata.com/influxdb/v1.0/administration/config/, the file doesn't exist on my machine after installing InfluxDB, so I assumed it was a typo.

Maybe worth mentioning this file on that page?

PR #662, opened 2016-09-09
fix instructions on where to get binaries

beckettsean:


beckettsean:
@rkuchan @kostasb

kostasb:
Are 14-day Demo licenses available for Uno plan only? If so we should note it to set expectations.
Other than that LGTM.

beckettsean:
Demo licenses are available for all but Gigawatt size clusters.

Issue #658, opened 2016-09-09
Documentation typo in the "Downsampling and Data Retention" page


fabricedesre:
###### URL for relevant page?
https://docs.influxdata.com/influxdb/v1.0/guides/downsampling_and_retention/

In the `Results` section, instead of `name: cpu` and `name: downsampled_cpu` this should be `name: orders` and `name: downsampled_orders`

rkuchan:
Thank you, @fabricedesre! This will be fixed by  https://github.com/influxdata/docs.influxdata.com/commit/2c52a037764861bda5e74cd9eddc3eefa49bfdce.

Issue #522, opened 2016-07-06
FEI entry


beckettsean:
https://github.com/influxdata/influxdb/issues/6967

fill doesn't always have data

beckettsean:
also https://github.com/influxdata/influxdb/issues/6878 if we haven't already documented that

rkuchan:
Link to from `fill()` docs in data exploration.

rkuchan:
Actually, influxdata/influxdb#6967 isn't documented.

PR #666, opened 2016-09-12
Fix a couple issues

rkuchan:
* Removes `beta` from the hostnames in the Enterprise documentation
* Fixes the typo in the downsampling guide, https://github.com/influxdata/docs.influxdata.com/issues/658
* Adds an FEI about `fill(previous)` and links to that FEI from the data exploration page, https://github.com/influxdata/docs.influxdata.com/issues/522


beckettsean:
@rkuchan we found and fixed the same `beta` issue at the same time. 

This looks great!

PR #667, opened 2016-09-12
remove beta, improve troubleshooting

beckettsean:
@rkuchan

PR #669, opened 2016-09-12
add verification highlights

beckettsean:
@rkuchan I cherry-picked the changes from https://github.com/influxdata/docs.influxdata.com/pull/667 to here, plus added a new `> **Verify:** ` pattern to highlight the double-checking steps in the install.

### DIFF
--- a/content/enterprise/v1.0/introduction/installation.md
+++ b/content/enterprise/v1.0/introduction/installation.md
@@ -53,18 +53,18 @@ file (the hostnames are representative):
 <Data_2_IP> enterprise-data-02
 ```
 
-> **Verify:** Before proceeding with the installation, verify on each server that the other 
+> **Verification steps:**
+>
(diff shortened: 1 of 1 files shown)


rkuchan:
@beckettsean, my last commit is mostly formatting, I:

* Changed `verify` to `verification steps`
* Reformatted the verfication steps notes - the codeblocks within `>` can be super finicky
* Changed the organzation of the `Join nodes to the cluster section` so that it has only two steps with verfication sections instead of four steps
* Removed the `verfication steps` title from any expected outputs - it makes the page less crowded and the expected outputs don't require any additional work by the user

Lastly, I reformatted the postgresql section because the codeblocks in the `>` weren't rendering well on the live site and I couldn't leave it alone.

rkuchan:
I think this is good to go!

beckettsean:
Awesome, thanks!

PR #672, opened 2016-09-13
Add doc for generating and using config file for Chronograf

aquarhead:
Explain `-sample-config` and `-config` like InfluxDB and Telegraf's doc.


beckettsean:
@willpiers @rossmcdonald @toddboom feedback?

Issue #668, opened 2016-09-12
Update series definition


rkuchan:
###### URL for relevant page?

At least:
- [x] https://docs.influxdata.com/influxdb/v1.0/guides/writing_data/
- [x] https://docs.influxdata.com/influxdb/v1.0/concepts/key_concepts/
- [x] https://docs.influxdata.com/influxdb/v1.0/concepts/glossary/
- [x] https://docs.influxdata.com/influxdb/v1.0/troubleshooting/frequently_encountered_issues/
- [x] https://docs.influxdata.com/influxdb/v1.0/query_language/data_exploration/
- [x] https://docs.influxdata.com/influxdb/v1.0/query_language/schema_exploration/
- [x] https://docs.influxdata.com/influxdb/v1.0/query_language/database_management/
- [x] https://docs.influxdata.com/influxdb/v1.0/query_language/functions/

rkuchan:
Things to pay attention to:

* series definition
* point definition
* writing duplicate points
* queries for series cardinality
* show series

rkuchan:
Will be fixed by: https://github.com/influxdata/docs.influxdata.com/pull/675

PR #675, opened 2016-09-13
Updates docs with the new definition of a series

rkuchan:
### Concepts/definitions

This PR updates the definitions for:
- series (it now includes field key)
- points
- points per second
- values per second 

(I believe points per second and values per second are essentially the same now)

The PR also introduces the concept of metaseries - the measurement-tag-set combination that we previously called series.
### Query description updates

`DROP SERIES` drops metaseries instead of series and `SHOW SERIES` shows metaseries instead of series.

`SLIMIT` limits metaseries, not series.
### Issues
- The PR removes the queries for series cardinality from the Frequently Encountered Issues page. If anyone knows how to query `_internal` to get the actual series cardinality let me know. The `database` measurement that we were using initially only has data on `numMeasurements` and `numSeries` (where `numSeries` does not include field keys) and I'm not sure how to get the number of field keys in there.
- There's no transition documentation that lets users know about the series definition change. Any ideas about how/where to put this are appreciated.

Fixes https://github.com/influxdata/docs.influxdata.com/issues/668

### DIFF
--- a/content/influxdb/v1.0/concepts/glossary.md
+++ b/content/influxdb/v1.0/concepts/glossary.md
@@ -42,17 +42,21 @@ Related entries: [retention policy](/influxdb/v1.0/concepts/glossary/#retention-
 
 ## field
 The key-value pair in InfluxDB's data structure that records metadata and the actual data value.
-Fields are required in InfluxDB's data structure and they are not indexed - queries on field values scan all points that match the specified time range and, as a result, are not performant relative to tags.
-
-*Query tip:* Compare fields to tags; tags are indexed.
+Fields are required in InfluxDB's data structure.
 
 Related entries: [field key](/influxdb/v1.0/concepts/glossary/#field-key), [field set](/influxdb/v1.0/concepts/glossary/#field-set), [field value](/influxdb/v1.0/concepts/glossary/#field-value), [tag](/influxdb/v1.0/concepts/glossary/#tag)
 
 ## field key
 The key part of the key-value pair that makes up a field.
 Field keys are strings and they store metadata.
+Field keys are indexed.
 
-Related entries: [field](/influxdb/v1.0/concepts/glossary/#field), [field set](/influxdb/v1.0/concepts/glossary/#field-set), [field value](/influxdb/v1.0/concepts/glossary/#field-value), [tag key](/influxdb/v1.0/concepts/glossary/#tag-key)
+Related entries:
+[field](#field),
+[field set](#field_set),
+[field value](#field-value),
+[series](#series),
+[tag key](#tag-key)
 
 ## field set
 The collection of field keys and field values on a point.
@@ -64,7 +68,7 @@ The value part of the key-value pair that makes up a field.
 Field values are the actual data; they can be strings, floats, integers, or booleans.
 A field value is always associated with a timestamp.
 
-Field values are not indexed - queries on field values scan all points that match the specified time range and, as a result, are not performant.
+Field values are not indexed - queries on field values scan all points that match the specified time range and, as a result, are not as performant as queries on tag values.
 
 *Query tip:* Compare field values to tag values; tag values are indexed.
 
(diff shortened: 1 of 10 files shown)


content/influxdb/v1.0/concepts/glossary.md
```diff
@@ -102,20 +136,16 @@ An independent `influxd` process.
...
+When you write a new point to the same series with the same timestamp, InfluxDB overwrites the existing the point with the new data.
 
 Related entries: [field set](/influxdb/v1.0/concepts/glossary/#field-set), [series](/influxdb/v1.0/concepts/glossary/#series), [timestamp](/influxdb/v1.0/concepts/glossary/#timestamp)
 
 ## points per second
-A deprecated measurement of the rate at which data are persisted to InfluxDB.
-The schema allows and even encourages the recording of multiple metric vales per point, rendering points per second ambiguous.
-
-Write speeds are generally quoted in values per second, a more precise metric.
+The preferred measurement of the rate at which data are persisted to InfluxDB.
```
beckettsean: I think this is wrong. Points per second records just 1 point for both of the following:

insert foo value=1
insert foo v=1,a=2,l=3,u=4,e=5

The latter inserts 5 values to the former's 1 value. We want to use values per second as our preferred metric to show how much actual data is persisting.

content/influxdb/v1.0/concepts/glossary.md
```diff
@@ -102,20 +136,16 @@ An independent `influxd` process.
...
+Write speeds are generally quoted in points per second.
```
beckettsean: I've been trying to move to vps, and I know that's what Paul wants.

content/influxdb/v1.0/concepts/glossary.md
```diff
@@ -158,51 +188,78 @@ See [InfluxQL Functions](/influxdb/v1.0/query_language/functions/#selectors) for
...
-| :-------------------- | :----- |
-| lorr@influxdata.com | start  |
-| lorr@influxdata.com | finish |
-| marv@influxdata.com     | start  |
-| marv@influxdata.com     | finish |
-| cliff@influxdata.com | start  |
-| cliff@influxdata.com | finish |
+The number of unique database, measurement, tag set, and field key combinations in an InfluxDB instance.
+
+For example, assume that an InfluxDB instance has a single datbase and one measurement.
```
beckettsean: database

content/influxdb/v1.0/concepts/glossary.md
```diff
@@ -291,9 +352,7 @@ When authentication is enabled, InfluxDB only executes HTTP requests that are se
 See [Authentication and Authorization](/influxdb/v1.0/administration/authentication_and_authorization/).
 
 ## values per second
-The preferred measurement of the rate at which data are persisted to InfluxDB. Write speeds are generally quoted in values per second.
-
-To calculate the values per second rate, multiply the number of points written per second by the number of values stored per point. For example, if the points have four fields each, and a batch of 5000 points is written 10 times per second, then the values per second rate is `4 field values per point * 5000 points per batch * 10 batches per second = 200,000 values per second`.
+See points per second.
```
beckettsean: See my earlier note about pps.

content/influxdb/v1.0/troubleshooting/frequently_encountered_issues.md
```diff
@@ -323,56 +306,66 @@ Writes an integer: `value=100i`
 Writes a float: `value=100`
 
 ## Writing duplicate points
-A point is uniquely identified by the measurement name, [tag set](/influxdb/v1.0/concepts/glossary/#tag-set), and timestamp.
-If you submit a new point with the same measurement, tag set, and timestamp as an existing point, the field set becomes the union of the old field set and the new field set, where any ties go to the new field set.
+A point is uniquely identified by the measurement name, [tag set](/influxdb/v1.0/concepts/glossary/#tag-set), [field key](/influxdb/v1.0/concepts/glossary/#field-key), and timestamp.
```
beckettsean: I'm not sure this is true. I think this is still one point:

insert foo value=1,foobar=2

Two values, one point. By this definition that would be two distinct points.

content/influxdb/v1.0/troubleshooting/frequently_encountered_issues.md
```diff
@@ -281,22 +280,6 @@ time			               count
 1970-01-01T00:00:00Z	 8
 ```
 
-## Querying for series cardinality
-
-The following queries return [series cardinality](/influxdb/v1.0/concepts/glossary/#series-cardinality):
-
-#### Series cardinality per database:
-```
-SELECT numSeries FROM "_internal".."database" GROUP BY "database" ORDER BY desc LIMIT 1
```
beckettsean: These are valuable queries that I'm not sure we have documented anywhere else. I would prefer not to remove them from the docs. Is there maybe a better place for them to live?

content/influxdb/v1.0/troubleshooting/frequently_encountered_issues.md
```diff
@@ -281,22 +280,6 @@ time			               count
...
-```
-#### Series cardinality across all database:
-```
-SELECT sum(numSeries) AS “total_series" FROM “_internal".."database" WHERE time > now() - 10s
```
beckettsean: These are valuable queries that I'm not sure we have documented anywhere else. I would prefer not to remove them from the docs. Is there maybe a better place for them to live?

content/influxdb/v1.0/write_protocols/line_protocol_reference.md
```diff
@@ -229,10 +229,9 @@ write Line Protocol to the database.
 
 ### Duplicate points
 
-A point is uniquely identified by the measurement name, tag set, and timestamp.
-If you submit Line Protocol with the same measurement, tag set, and timestamp,
-but with a different field set, the field set becomes the union of the old
-field set and the new field set, where any conflicts favor the new field set.
+A point is uniquely identified by the measurement name, tag set, field key, and timestamp.
```
beckettsean: Same question about what is a point. We need to resolve this.

content/influxdb/v1.0/write_protocols/line_protocol_tutorial.md
```diff
@@ -413,10 +413,9 @@ on the [HTTP API](/influxdb/v1.0/tools/api/#write), the
 
 ### Duplicate points
 
-A point is uniquely identified by the measurement name, tag set, and timestamp.
-If you submit Line Protocol with the same measurement, tag set, and timestamp,
-but with a different field set, the field set becomes the union of the old
-field set and the new field set, where any conflicts favor the new field set.
+A point is uniquely identified by the measurement name, tag set, field key, and timestamp.
```
beckettsean: same question about points

PR #686, opened 2016-09-15
Fixes #660, better "if" function docs

### LINKED ISSUE #660: Need more Documentation on Conditional Functions
###### URL
https://docs.influxdata.com/kapacitor/v1.0/tick/expr/#conditional-functions

I want a more complex lambda expressions using AND condition. Looks like conditional functions will fit to the need, but need more info on it.

nathanielc:


nathanielc:
@beckettsean @sarathj How does this look?

PR #696, opened 2016-09-19
Wait for metastore to load OSS data

kostasb:
Shards from an OSS instance will take some time before they are mapped in the meta cluster.
Also added influxdctl-show as a final verification step.

@rkuchan @beckettsean


rkuchan:
👍

PR #699, opened 2016-09-19
Add clustering doc

### LINKED ISSUE #540: new page: technical doc on how clustering works
We need to say something about Raft, meta nodes, data nodes, etc.

pauldix:
Fixes #540 

@jwilder @beckettsean


content/enterprise/v1.0/concepts/clustering.md
```diff
@@ -0,0 +1,123 @@
...
+* `any` - return success to the client as soon as any server has responded or server D has written to the local disk queue
+* `one` - return success to the client when either server A or B has accepted the write
+* `quorum` - return success when both A and B return success. This option only useful is replication factors that are greater than 2.
+* `all` - return success only when both A and B return success.
+
+The important thing to note is how failures are handled. In the case of failures, the database will use the hinted handoff system.
+
+### Hinted Handoff
+
+Hinted handoff is how InfluxEnterprise deals with data node failures while writes are happening. Hinted handoff is essentially a durable disk based queue. In the case of `any`, `one` or `quorum` writes, hinted handoff will be used if one or more replicas return an error after a success has already been returned to the client.
```
beckettsean: Hinted handoff is how InfluxEnterprise deals with data node outages while writes are happening. Hinted handoff is essentially a durable disk based queue. When writing at `any`, `one` or `quorum` consistency, hinted handoff is used when one or more replicas return an error after a success has already been returned to the client. When writing at `all` consistency, writes cannot return success unless all nodes return success, so hinted handoff is not used. 

Let's again use the example of a write coming to `D` that should go to shard `1` on `A` and `B`. If we specified a consistency level of `one` and node `A` returns success, `D` will immediately return success to the client even though the write to `B` is still in progress.

Now let's assume that `B` returns an error. Node `D` then puts the write into its hinted handoff queue for shard `1` on node `B`. In the background, node `D` will continue to attempt to empty the hinted handoff queue by writing the data to node `B`. The configuration file has settings for the maximum size and age of data in hinted handoff queues.

If a data node is restarted it will check for pending writes in the hinted handoff queues and resume attempts to replicate the writes. The important thing to note is that the hinted handoff queue is durable and does survive a process restart.

When restarting nodes within an active cluster, during upgrades or maintenance, for example, other nodes in the cluster will store hinted handoff writes to the offline node and replicate them when the node is again available. Thus, a healthy cluster should have enough resource headroom on each data node to handle the burst of hinted handoff writes following a node outage. The returning node will need to handle both the steady state traffic and the queued hinted handoff writes from other nodes, meaning its write traffic will have a significant spike following any outage of more than a few seconds.

content/enterprise/v1.0/concepts/clustering.md
```diff
@@ -0,0 +1,123 @@
...
+
+If server D is restarted it will check if there are any writes sitting in hinted handoff and continue to attempt to replicate those. The important thing to note is that the hinted handoff queue is durable and will survive server restarts.
+
+When doing restarts of servers within an active cluster it's expected that other servers will store hinted handoff writes and replicate them when the server comes back up. Because of this, it's best to operate a cluster at less than full 100% utilization. A healthy cluster should have headroom to handle burst write and query traffic or recovery from temporary outages in any part of the cluster.
+
+## Queries in a Cluster
+
+Queries in a cluster are distributed based on the time range being queried and replication factor. For example if we have a replication factor of 4 and shard durations of one day, for each day the query hits, any one of the 4 servers that stores a replica of that shard can be queried.
```
beckettsean: Queries in a cluster are distributed based on the time range being queried and the replication factor of the data. For example if the retention policy has a replication factor of 4 and shard durations of one day, for each day of time covered by a query, the node receiving the query randomly picks any of the 4 servers that store a replica of that shard to receive the query.

beckettsean: I think the last sentence is correct but can someone verify, please?

content/enterprise/v1.0/concepts/clustering.md
beckettsean: > When writing at all consistency, writes cannot return success unless all nodes return success, so hinted handoff is not used.

Is this true? It seems to be from my understanding, but doesn't that also mean that consistency `all` clusters can't handle will return write failures for all writes if any receiving data node is down? 

If it's not true, then why was `all` excluded from the previous paragraph?

content/enterprise/v1.0/concepts/clustering.md
```diff
@@ -0,0 +1,123 @@
...
+
+A query that hits multiple shard groups (i.e. days) will run those individual shard queries in parallel while fanning out to the other servers in the cluster that must be hit. As the results come in from each shard, they will be combined together to form the final result that gets returned to the user.
```
beckettsean: What's this paragraph trying to express that the previous one doesn't? I don't quite grok what it means to `run those individual shard queries in parallel` while also `fanning out to the other servers in the cluster`. How do those differ? In both cases, I view that the receiving node is forwarding the query to one node for every queried shard.

beckettsean:
@pauldix I'm also happy to make direct commits to your branch if you prefer that.

@rkuchan do you have suggestions?

content/enterprise/v1.0/concepts/clustering.md
```diff
@@ -0,0 +1,142 @@
...
+
+The important thing to note is how failures are handled. In the case of failures, the database will use the hinted handoff system.
+
+### Hinted Handoff
+
+Hinted handoff is how InfluxEnterprise deals with data node outages while writes are happening. Hinted handoff is essentially a durable disk based queue. When writing at `any`, `one` or `quorum` consistency, hinted handoff is used when one or more replicas return an error after a success has already been returned to the client. When writing at `all` consistency, writes cannot return success unless all nodes return success, so hinted handoff is not used.
```
beckettsean: > When writing at all consistency, writes cannot return success unless all nodes return success, so hinted handoff is not used.

Is this true? It seems to be from my understanding, but doesn't that also mean that consistency all clusters can't handle will return write failures for all writes if any receiving data node is down?

If it's not true, then why was `all` excluded from the previous paragraph?

content/enterprise/v1.0/concepts/clustering.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Queries in a Cluster
+
+Queries in a cluster are distributed based on the time range being queried and the replication factor of the data. For example if the retention policy has a replication factor of 4 and shard durations of one day, for each day of time covered by a query, the node receiving the query randomly picks any of the 4 servers that store a replica of that shard to receive the query.
+
+A query that hits multiple shard groups (i.e. days) will run those individual shard queries in parallel while fanning out to the other servers in the cluster that must be hit. As the results come in from each shard, they will be combined together to form the final result that gets returned to the user.
```
beckettsean: What's this paragraph trying to express that the previous one doesn't? I don't quite grok what it means to `run those individual shard queries in parallel` while also `fanning out to the other servers in the cluster`. How do those differ? In both cases, I view that the receiving node is forwarding the query to one node for every queried shard.

content/enterprise/v1.0/concepts/clustering.md
```diff
@@ -0,0 +1,142 @@
...
+
+When restarting nodes within an active cluster, during upgrades or maintenance, for example, other nodes in the cluster will store hinted handoff writes to the offline node and replicate them when the node is again available. Thus, a healthy cluster should have enough resource headroom on each data node to handle the burst of hinted handoff writes following a node outage. The returning node will need to handle both the steady state traffic and the queued hinted handoff writes from other nodes, meaning its write traffic will have a significant spike following any outage of more than a few seconds.
+
+## Queries in a Cluster
+
+Queries in a cluster are distributed based on the time range being queried and the replication factor of the data. For example if the retention policy has a replication factor of 4 and shard durations of one day, for each day of time covered by a query, the node receiving the query randomly picks any of the 4 servers that store a replica of that shard to receive the query.
```
beckettsean: I think the last sentence is correct but can someone verify, please?

content/enterprise/v1.0/concepts/clustering.md
beckettsean: It is true, per convo with @pauldix

content/enterprise/v1.0/concepts/clustering.md
pauldix: It is. That will change over time to include things like health of the node and load.

content/enterprise/v1.0/concepts/clustering.md
pauldix: It's trying to capture that queries that hit multiple shards will be executed in parallel (as opposed to serially). So they'll potentially hit many servers in the cluster at the same time. In a fully replicated cluster, queries always opt for hitting the local store. Even then the query operations on each shard are run in parallel.

content/enterprise/v1.0/concepts/clustering.md
jwilder: Hinted handoff is still used even with `all` and a write fails.  We won't return a successful write response to the caller, but the write will still be queued and retried later via hinted handoff.

beckettsean: @jwilder I rewrote the end of that paragraph. Can you check it again?

content/enterprise/v1.0/concepts/clustering.md
beckettsean: @pauldix I re-wrote the final two paragraphs. Can you check them again, please?

content/enterprise/v1.0/concepts/clustering.md
pauldix: @jwilder that's unexpected for me. I would have thought that it wouldn't use HH

PR #707, opened 2016-09-22
Vertical tab widget

wfro:
### Overview

Adds a simple vertical tabs widget, where authors can specify a list of
tabs and their corresponding content, and Hugo will render a nice tab
widget where users can switch between tabs (which saves a lot of
vertical space on the docs). Full credit to @rkuchan.

It's intended usage is for code examples, though it's meant to be
general purpose and could work for any content you'd want to tabularize.

### How to use it

Disclaimer: I'm not an expert at Hugo and the interface to actually use
this widget isn't great. Really all it does is provide a bunch of Hugo
[shortcodes](https://gohugo.io/extras/shortcodes#creating-your-own-shortcodes),
which just insert some raw HTML into the markdown.

A few shortcodes are exposed:
* `vertical-tabs` -- this is the main container and needs to wrap all
  the things
* `tabs` -- define all tab titles here, using markdown links (e.g. [Tab 1](#)).
  The links don't need a proper href, but having an `a` tag is required
  for styles to work
* `tab-content-container` -- another container, but this time you'll
  need to wrap all of your content definitions with this shortcode.
* `tab-content` -- this wraps each content definition (so if you had
  two tabs, you'd defined two `tab-content` blocks.

A more complete example:

> Note that the `tab-content` blocks should be proper markdown code blocks, but I removed the backticks to make sure this github comment rendered correctly :D

```
{{< vertical-tabs >}}
{{% tabs %}}
[With SELECT](#)
[With SELECT and INTO](#)
[Create a database](#)
{{% /tabs %}}
{{< tab-content-container >}}
{{% tab-content %}}
Query data with a `SELECT` statement:
$ curl -GET 'http://localhost:8086/query?db=mydb' --data-urlencode
'q=SELECT * FROM "mymeas"'

{"results":[{"series":[{"name":"mymeas","columns":["time","myfield","mytag1","mytag2"],"values":[["2016-05-20T21:30:00Z",12,"1",null],["2016-05-20T21:30:20Z",11,"2",null],["2016-05-20T21:30:40Z",18,null,"1"],["2016-05-20T21:31:00Z",19,null,"3"]]}]}]}
{{% /tab-content %}}

{{% tab-content %}}
Query data with a `SELECT` statement and an `INTO` clause:
$ curl -XPOST 'http://localhost:8086/query?db=mydb' --data-urlencode
'q=SELECT * INTO "newmeas" FROM "mymeas"'

{"results":[{"series":[{"name":"result","columns":["time","written"],"values":[["1970-01-01T00:00:00Z",4]]}]}]}
{{% /tab-content %}}

{{% tab-content %}}
Create a database:
$ curl -XPOST 'http://localhost:8086/query' --data-urlencode 'q=CREATE
DATABASE "mydb"'

{"results":[{}]}
{{% /tab-content %}}
{{< /tab-content-container >}}
{{< /vertical-tabs >}}
```

One gotcha: the order in which you define tabs and content matters, so if you
define tab1 first, make sure you define content1 first.

The general idea is that using something like `{{% tab }}` is more pleasant to
work with than `<div class="vertical-tab">`. Ideally, authors would only have to
specify a list of tabs, and some content because those are the only
dynamic pieces. But because we're working directly in markdown, I
didn't immediately see an easy way to define a partial template like that.


beckettsean:
@toddboom @desa @121watts can one of y'all to review this docs PR?

121watts:
@beckettsean for now this is a thumbs down.  I installed and attempted to run locally.  Looks like the javascript is working just fine, but the styles aren't appearing to be applied.  I've let @alexpaxton know what's going on and will work with him on this when he's done being in an airplane.

121watts:
@beckettsean this PR is good to go! Merge away.

beckettsean:
awesome, thanks, @121watts!

PR #719, opened 2016-09-26
AuthN and AuthZ links broken due to administration / query_language cross indexing

andyfeller:
It seems all of the authentication and authorization links are broken as the content actually lives within query_language but links reference administration.  Assuming this content is meant to live within query_language, I've updated links to point there to get this working.


andyfeller:
@rkuchan you might be interested in this.

rkuchan:
About to push a fix for this. Thanks, @andyfeller.

PR #725, opened 2016-09-27
Typo InfluxDB- v1.0- concepts- storage engine- compactions

peeyush-tm:
Typo in [influxdb v1.0 storage engine's concept compactions](https://docs.influxdata.com/influxdb/v1.0/concepts/storage_engine/#compactions)
sub-section - Index Optimization (Line Number 3)
Before an index ~~iptimization~~, changed to, Before an index optimization


beckettsean:
Thanks, @peeyush-tm!

PR #745, opened 2016-09-30
alphabetize queries in spec

beckettsean:
@rkuchan noticed today that the statements were no longer in alphabetical order. Thought I'd fix that.

### DIFF
--- a/content/influxdb/v1.0/query_language/spec.md
+++ b/content/influxdb/v1.0/query_language/spec.md
@@ -595,36 +595,6 @@ SELECT mean("value") FROM "cpu" WHERE "region" = 'uswest' GROUP BY time(10m) fil
 SELECT mean("value") INTO "cpu_1h".:MEASUREMENT FROM /cpu.*/
 ```
 
-## Clauses
-
-```
-from_clause     = "FROM" measurements .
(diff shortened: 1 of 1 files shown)


content/influxdb/v1.0/query_language/spec.md
```diff
@@ -561,6 +561,70 @@ KILL QUERY 36
...
+
+```sql
+-- select mean value from the cpu measurement where region = 'uswest' grouped by 10 minute intervals
+SELECT mean("value") FROM "cpu" WHERE "region" = 'uswest' GROUP BY time(10m) fill(0)
+
+-- select from all measurements beginning with cpu into the same measurement name in the cpu_1h retention policy
+SELECT mean("value") INTO "cpu_1h".:MEASUREMENT FROM /cpu.*/
+```
+
+## Clauses
```
rkuchan: Is this section supposed to be after all of the statements? Fine if this was intentional.

beckettsean: Not intentional, thanks for catching. I moved it back to the end of all the statements.

rkuchan:
👍

PR #746, opened 2016-10-01
Mention error condition of CREATE RETENTION POLICY

mark-rushakoff:


beckettsean:
Thanks, @mark-rushakoff, an important distinction!

PR #753, opened 2016-10-03
clarify hinted handoff

### LINKED ISSUE #750: Detail errors in hinted handoff
https://docs.influxdata.com/enterprise/v1.0/concepts/clustering#hinted-handoff

Rough outline of extra info to add:

#### `hinted handoff queue not empty`

The node handling the write already has data in HH for the destination node. New data is not sent until existing data in HH is flushed to the destination node, so this error indicates that the queue is still depleting.

#### `queue is full`

The node handling the write cannot store the incoming write into the HH queue for the destination node because it would exceed the maximum size of the HH queue (should be `hh.max-size` in the toml config). This error typically indicates a catastrophic condition for the cluster - one node may have been offline or unable to accept writes for an excessive duration.

----

In both these cases (and other HH error cases which should not occur during normal operation), the InfluxDB server returns a 500 status code. Some HTTP clients can be configured to retry requests if a server returns a 500 (e.g. `curl --retry`). If a request is retried due to a HH error, the system will add duplicate points to HH queues, potentially causing the system to stay in "HH recovery" state even longer.

beckettsean:
fixes #750


beckettsean:
@rkuchan @mark-rushakoff @jwilder

content/enterprise/v1.0/troubleshooting/frequently_asked_questions.md
```diff
@@ -46,3 +48,15 @@ Log output for the data node that accepts the lease:
...
+
+Note that for some [write consistency](/enterprise/v1.0/concepts/clustering/#write-consistency) settings, InfluxDB may return a write error (500) for the write attempt, even if the points are successfully queued in hinted handoff. Some write clients may attempt to resend those points, leading to duplicate points being added to the hinted handoff queue and lengthening the time it takes for the queue to drain. If the queues are not draining, consider temporarily downgrading the write consistency setting, or pause retries on the write clients until the hinted handoff queues full drain.
```
rkuchan: full drain -> fully drain

rkuchan:
one small fix, otherwise: 👍

beckettsean:
Thanks for the reviews!

PR #755, opened 2016-10-04
Add vertical tab widget to InfluxDB and Telegraf

rkuchan:
Integrates the vertical tab widget with the installation pages for InfluxDB and Telegraf. Example:

image


rkuchan:
@kostasb, I had to `gulp less` to get the formatting to show up properly - as a result, I have some changes to `/static/` files that I didn't commit. Should I push those up as well or will the build process take care of it?

kostasb:
@rkuchan The staging website doesn't show up correctly, so I guess it needs the new static artifacts to be committed as well. The build process does not involve npm so it will not detect the changes.

Here is the staging site for this branch:
http://docs.staging.influxdata.com:31003/influxdb/v1.0/introduction/installation/#installation

Issue #772, opened 2016-10-06
Kapacitor tasks define command needs to fix with retention policy as 'autogen'


manojgour:
###### URL for relevant page?
https://docs.influxdata.com/kapacitor/v1.0/introduction/getting_started/

###### What products and version are you using?
Kapacitor V1.0
###### Where did you look before opening the issue?
Checked few previous posts here on Github; got some clue with dbrp is missing then I did 'SHOW SUBSCRIPTION' that's where I found retention policy is no longer "default"

Kapacitor command for defining tasks is not working as described in document.

Below command was working fine for earlier version i.e v0.13:
kapacitor define cpu_alert -type stream -tick cpu_alert.tick -dbrp kapacitor_example.default

With Kapacitor new version, when I changed the retention policy to 'autogen' it worked fine:
kapacitor define cpu_alert -type stream -tick cpu_alert.tick -dbrp kapacitor_example.autogen

rkuchan:
Thank you for catching that, @manojgour!

Issue #762, opened 2016-10-05
ALTER RP no longer resets the shard group duration


rkuchan:
For 1.0.2: https://github.com/influxdata/influxdb/issues/7150

FAQ on rps, data, shard durations.

Issue #680, opened 2016-09-14
Documentation InfluxDB 1.0: "401 Authorization"  changed to "403 Forbidden"


PaulKuiper:
https://docs.influxdata.com/influxdb/v1.0/query_language/authentication_and_authorization/#authentication-and-authorization-http-errors

Al the way at the bottom in the section: Authentication and authorization HTTP errors

It states that a "HTTP 401 authorization" error is returned. 
However in https://github.com/influxdata/influxdb/pull/6978  this was changed to "403 Forbidden"

rkuchan:
Thank you, @PaulKuiper!

Issue #476, opened 2016-05-27
Add entry to FEI page about mixing aggregates and non-aggregates


rkuchan:
###### URL for relevant page?
https://docs.influxdata.com/influxdb/v0.13/troubleshooting/frequently_encountered_issues/

###### What products and version are you using?
all

###### Where did you look before opening the issue?
NA

PR #774, opened 2016-10-07
Docs issues

rkuchan:
Fixes issues:

* #762: Update the `Why aren't data dropped after I've altered a retention policy?` FAQ for 1.0.2 behavior.
* #522: Add documentation for `fill()` behavior with no data.
* #476: Add FAQ on mixing aggregates and non-aggregates.
* #772: Remove mentions of the `default` RP and replace with `autogen` in the 1.0 Kapacitor docs.
* #680: Update description of auth errors to include `403`.

### DIFF
--- a/content/influxdb/v1.0/troubleshooting/frequently-asked-questions.md
+++ b/content/influxdb/v1.0/troubleshooting/frequently-asked-questions.md
@@ -129,15 +129,38 @@ statement.
 
 ## Why aren't data dropped after I've altered a retention policy?
 
-By default, InfluxDB checks to enforce a RP every 30 minutes.
+Several factors explain why data may not be immediately dropped after a
+retention policy (RP) change.
+
+The first and most likely cause is that, by default, InfluxDB checks to enforce
+an RP every 30 minutes.
 You may need to wait for the next RP check for InfluxDB to drop data that are
 outside the RP's new `DURATION` setting.
 The 30 minute interval is
 [configurable](/influxdb/v1.0/administration/config/#check-interval-30m0s).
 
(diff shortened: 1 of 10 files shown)


rkuchan:
@beckettsean, ready for review.

content/influxdb/v1.0/troubleshooting/frequently-asked-questions.md
```diff
@@ -127,24 +128,16 @@ Check your retention policy's shard group duration with the
...
 
-First, by default, InfluxDB checks to enforce a RP every 30 minutes.
+By default, InfluxDB checks to enforce a RP every 30 minutes.
 You may need to wait for the next RP check for InfluxDB to drop data that are
 outside the RP's new `DURATION` setting.
 The 30 minute interval is
 [configurable](/influxdb/v1.0/administration/config/#check-interval-30m0s).
 
-Second,
-changing an RP's `DURATION` [may cause](#what-is-the-relationship-between-shard-group-durations-and-retention-policies)
```
beckettsean: This info is still true, isn't it? We don't subdivide already-made shards if someone changes the RP.

Is this the one you moved to its own question?

rkuchan: I was only thinking about `DURATION` changes no longer automatically resetting the `SHARD DURATION` (https://github.com/influxdata/influxdb/issues/7150) which led to unexpected data retention. Now it'll return an error if `DURATION` is less than the shard duration.

Having some trouble thinking through this but will the inability to subdivide shards still cause unexpected data retention?

beckettsean:
last update looks good, Regan, thanks!

PR #776, opened 2016-10-07
Change default -> autogen in Kapacitor getting started guide

nathanielc:
This was causing some confusion since the doc was out of date with 1.0 defaults.

See https://github.com/influxdata/kapacitor/issues/960


beckettsean:
👍

PR #777, opened 2016-10-07
Document internal stats

mark-rushakoff:
First draft of documenting internal stats.

Ready for review and feedback.

### DIFF
--- a/content/enterprise/v1.0/troubleshooting/internal_statistics.md
+++ b/content/enterprise/v1.0/troubleshooting/internal_statistics.md
@@ -19,6 +19,11 @@ a few examples showing good and problematic query responses for some or all of t
 These are the metrics that are unique to InfluxEnterprise. 
 For metrics shared by both InfluxDB and InfluxEnterprise, please refer to the [InfluxDB `_internal` documentation](/influxdb/v1.0/troubleshooting/internal_statistics).
 
+In a cluster, each measurement in `_internal` has three tags:
+* `clusterID` - the UUID of the cluster recording the `_internal` metrics
+* `hostname` - the hostname of the node reporting the metrics
+* `nodeID` - the hostname and port of the node reporting the metrics
+
 ## cluster
 
 The `cluster` measurement tracks statistics related to the clustering features of the data nodes in InfluxEnterprise.
--- a/content/enterprise/v1.0/troubleshooting/internal_statistics.md
+++ b/content/enterprise/v1.0/troubleshooting/internal_statistics.md
@@ -104,7 +104,9 @@ Subsequent write requests to this node, destined for the same remote node, do no
 ## hh_processor
 
 The `hh_processor` tracks statistics about hinted handoff processors, one for each remote node.
-The tags indicate the destination node.
+The `hh_processor` measurement has two additional tags:
+* `node` - the destination node for the recorded metrics
+* `path` - the path to the durable hinted handoff queue on disk
 
 | fieldKey | fieldType | units |
 | :---- | :---- | :---- |
--- a/content/enterprise/v1.0/troubleshooting/internal_statistics.md
+++ b/content/enterprise/v1.0/troubleshooting/internal_statistics.md
@@ -50,11 +50,11 @@ For metrics shared by both InfluxDB and InfluxEnterprise, please refer to the [I
 
 ## cq  
 
-See \<link to influxdb doc\>
+See [InfluxDB `_internal` documentation](/influxdb/v1.0/troubleshooting/internal_statistics).
 
 ## database  
 
-See \<link to influxdb doc\>
+See [InfluxDB `_internal` documentation](/influxdb/v1.0/troubleshooting/internal_statistics).
 
 ## hh
 
@@ -98,39 +98,39 @@ See \<link to influxdb doc\>
 
 ## httpd  
 
-See \<link to influxdb doc\>
+See [InfluxDB `_internal` documentation](/influxdb/v1.0/troubleshooting/internal_statistics).
 
 ## queryExecutor  
 
-See \<link to influxdb doc\>
+See [InfluxDB `_internal` documentation](/influxdb/v1.0/troubleshooting/internal_statistics).
 
 ## runtime  
 
-See \<link to influxdb doc\>
+See [InfluxDB `_internal` documentation](/influxdb/v1.0/troubleshooting/internal_statistics).
 
 ## shard  
 
-See \<link to influxdb doc\>
+See [InfluxDB `_internal` documentation](/influxdb/v1.0/troubleshooting/internal_statistics).
 
 ## subscriber  
 
-See \<link to influxdb doc\>
+See [InfluxDB `_internal` documentation](/influxdb/v1.0/troubleshooting/internal_statistics).
 
 ## tsm1_cache  
 
-See \<link to influxdb doc\>
+See [InfluxDB `_internal` documentation](/influxdb/v1.0/troubleshooting/internal_statistics).
 
 ## tsm1_engine  
 
-See \<link to influxdb doc\>
+See [InfluxDB `_internal` documentation](/influxdb/v1.0/troubleshooting/internal_statistics).
 
 ## tsm1_filestore  
 
-See \<link to influxdb doc\>
+See [InfluxDB `_internal` documentation](/influxdb/v1.0/troubleshooting/internal_statistics).
 
 ## tsm1_wal  
 
-See \<link to influxdb doc\>
+See [InfluxDB `_internal` documentation](/influxdb/v1.0/troubleshooting/internal_statistics).
 
 
 ## write  
--- a/content/enterprise/v1.0/troubleshooting/internal_statistics.md
+++ b/content/enterprise/v1.0/troubleshooting/internal_statistics.md
@@ -21,32 +21,51 @@ For metrics shared by both InfluxDB and InfluxEnterprise, please refer to the [I
 
 ## cluster
 
+The `cluster` measurement tracks statistics related to the clustering features of the data nodes in InfluxEnterprise.
+The tags on the series indicate the source host of the stat.
+
 | fieldKey | fieldType | units |
 | :---- | :---- | :---- |
-| copyShardReq | integer | bananas |
-| createIteratorReq | integer | bananas |
-| expandSourcesReq | integer | bananas |
-| fieldDimensionsReq | integer | bananas |
-| removeShardReq | integer | bananas |
-| writeShardFail | integer | bananas |
-| writeShardPointsReq | integer | bananas |
-| writeShardReq | integer | bananas |
+| copyShardReq | integer | none |
+| createIteratorReq | integer | none |
+| expandSourcesReq | integer | none |
+| fieldDimensionsReq | integer | none |
+| removeShardReq | integer | none |
+| writeShardFail | integer | none |
+| writeShardPointsReq | integer | none |
+| writeShardReq | integer | none |
 
 ###copyShardReq
 
+The `copyShardReq` statistic is incremented every time there is an internal request to copy a shard _from_ this data node to another data node.
+
 ###createIteratorReq
+
+The `createIteratorReq` statistic is incremented every time a remote node requests to remotely read data from this node.
 
 ###expandSourcesReq
 
+The `expandSourcesReq` statistic is incremented every time a remote node requests to find measurements on this node that match a given regular expression.
+
 ###fieldDimensionsReq
+
+The `fieldDimensionsReq` statistic is incremented every time a remote node requests information about the fields and associated types, and tag keys, of measurements on this node.
 
 ###removeShardReq
 
+The `removeShardReq` statistic is incremented every time there is an internal request to delete a shard from this data node.
+
 ###writeShardFail
+
+The `writeShardFail` statistic is incremented every time there is an internal write request from a remote node, and the write fails.
 
 ###writeShardPointsReq
 
+The `writeShardPointsReq` statistic is incremented for every point in every internal write request from any remote node.
+
 ###writeShardReq
+
+The `writeShardReq` statistic is incremented every time there is an internal write request from a remote node, regardless of success.
 
 ## cq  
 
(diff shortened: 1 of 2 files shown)


mark-rushakoff:
Upcoming stat changes for 1.1 that I'm aware of:

* [new `walPath` tag on shard stats](https://github.com/influxdata/influxdb/pull/7383)
* `hh_processor.queueBytes` is an absolute value that indicates how many bytes remain in the queue. The number should remain correct across restarts, unlike the `bytesRead` and `bytesWritten` fields.
* [active compaction stats](https://github.com/influxdata/influxdb/pull/7281), not yet merged into 1.1.

content/enterprise/v1.0/troubleshooting/internal_statistics.md
```diff
@@ -19,154 +19,242 @@ a few examples showing good and problematic query responses for some or all of t
...
+| subWriteDrop | integer | none |
+| subWriteOk | integer | none |
+| writeError | integer | none |
+| writeOk | integer | none |
+| writePartial | integer | none |
+| writeTimeout | integer | none |
 
 ###pointReq
 
+The `pointReq` statistic is incremented for every point requested to be written to this node.
```
beckettsean: whether successful or not? Like, if I send a bad point, is this incremented?

content/enterprise/v1.0/troubleshooting/internal_statistics.md
```diff
@@ -19,154 +19,242 @@ a few examples showing good and problematic query responses for some or all of t
...
+
 ###pointReqLocal
 
+The `pointReqLocal` statistic is incremented for every point received for write by this node and able to be written to a shard that exists on the same node.
+
 ###pointReqRemote
 
+The `pointReqRemote` statistic is incremented for every point received for write by this node but needed to be forwarded to a remote node.
+This statistic does not distinguish between requests that are directly written to the destination node versus enqueued in hinted handoff for the destination node.
+
```
beckettsean: Does that mean that `pointReqRemote` - `pointReqHH` = the number of points successfully written directly to the destination node (e.g. success to remote node that avoids HH)?

content/enterprise/v1.0/troubleshooting/internal_statistics.md
```diff
@@ -19,154 +19,242 @@ a few examples showing good and problematic query responses for some or all of t
...
 
+The `writeOk` statistic is incremented any time a batch of points is written at the requested consistency level.
+
 ###writePartial
 
+The `writePartial` statistic is incremented any time a batch of points is written to at least one node but did not meet the requested consistency level.
+
 ###writeTimeout
+
+The `writeTimeout` statistic is incremented any time a write failed to complete within the default write timeout duration.
```
beckettsean: a write to this node, or from this node?

content/enterprise/v1.0/troubleshooting/internal_statistics.md
```diff
@@ -21,152 +21,231 @@ For metrics shared by both InfluxDB and InfluxEnterprise, please refer to the [I
...
+| writeShardPointsReq | integer | none |
+| writeShardReq | integer | none |
 
 ###copyShardReq
 
+The `copyShardReq` statistic is incremented every time there is an internal request to copy a shard _from_ this data node to another data node.
+
 ###createIteratorReq
 
+The `createIteratorReq` statistic is incremented every time a remote node requests to remotely read data from this node.
```
mark-rushakoff: Yes to both questions. Meta nodes never request time series data from data nodes, AFAIK.

content/enterprise/v1.0/troubleshooting/internal_statistics.md
```diff
@@ -21,152 +21,231 @@ For metrics shared by both InfluxDB and InfluxEnterprise, please refer to the [I
...
 
+The `expandSourcesReq` statistic is incremented every time a remote node requests to find measurements on this node that match a given regular expression.
+
 ###fieldDimensionsReq
 
+The `fieldDimensionsReq` statistic is incremented every time a remote node requests information about the fields and associated types, and tag keys, of measurements on this node.
+
 ###removeShardReq
 
+The `removeShardReq` statistic is incremented every time there is an internal request to delete a shard from this data node.
```
mark-rushakoff: I think remove-shard would trigger it, but so would `drop database`, and potentially `drop series` or `drop measurement` if that statement removed all of the remaining data in a shard.

mark-rushakoff:
I started to respond to some of the questions from memory, but it's looking like I should review all of them along with the source code. I'll plan on getting to that likely early next week.

beckettsean:
Thanks, @mark-rushakoff! We can also ask the rest of the team for input, but if the code spelunking is enjoyable for you or is providing valuable code review then that method works, too.

content/enterprise/v1.0/troubleshooting/internal_statistics.md
```diff
@@ -19,154 +19,238 @@ a few examples showing good and problematic query responses for some or all of t
 These are the metrics that are unique to InfluxEnterprise. 
 For metrics shared by both InfluxDB and InfluxEnterprise, please refer to the [InfluxDB `_internal` documentation](/influxdb/v1.0/troubleshooting/internal_statistics).
 
+In a cluster, each measurement in `_internal` has three tags:
+* `clusterID` - the UUID of the cluster recording the `_internal` metrics
+* `hostname` - the hostname of the node reporting the metrics
```
mark-rushakoff: The `hostname` is added to all statistics and should indicate the hostname as reported by the operating system (set [here in open source](https://github.com/influxdata/influxdb/blob/ff307047057b7797418998a4ed709b0c0f346324/monitor/service.go#L395)).

`nodeID` is set in closed source, and it looks like it should be set to the hostname or bind address as set in the config file.

content/enterprise/v1.0/troubleshooting/internal_statistics.md
```diff
@@ -21,152 +21,231 @@ For metrics shared by both InfluxDB and InfluxEnterprise, please refer to the [I
...
 
+The `createIteratorReq` statistic is incremented every time a remote node requests to remotely read data from this node.
+
 ###expandSourcesReq
 
+The `expandSourcesReq` statistic is incremented every time a remote node requests to find measurements on this node that match a given regular expression.
+
 ###fieldDimensionsReq
 
+The `fieldDimensionsReq` statistic is incremented every time a remote node requests information about the fields and associated types, and tag keys, of measurements on this node.
```
mark-rushakoff: Upon closer review, it looks like field dimensions (fields and associated types, and tag keys) are retrieved for  every SELECT statement. `fieldDimensionsReq` means that the information had to be retrieved from a remote node.

content/enterprise/v1.0/troubleshooting/internal_statistics.md
```diff
@@ -21,152 +21,231 @@ For metrics shared by both InfluxDB and InfluxEnterprise, please refer to the [I
...
+
 ###removeShardReq
 
+The `removeShardReq` statistic is incremented every time there is an internal request to delete a shard from this data node.
+
 ###writeShardFail
 
+The `writeShardFail` statistic is incremented every time there is an internal write request from a remote node, and the write fails.
```
mark-rushakoff: It's the cousin of OSS shard stat `writeReqErr`. 

A write request over HTTP is received by Node A. Node A does not have the shard locally, so it creates an internal request to Node B instructing what to write and to which shard. If Node B sees the request and if anything goes wrong, Node B increments its own `writeShardFail`. Depending on what went wrong, in most circumstances Node B would also increment its `writeReqErr` stat inherited from OSS.

If Node A had the shard locally, there would be no internal request to write data to a remote node, so `writeShardFail` would not be incremented.

content/enterprise/v1.0/troubleshooting/internal_statistics.md
```diff
@@ -21,152 +21,231 @@ For metrics shared by both InfluxDB and InfluxEnterprise, please refer to the [I
...
+
 ###writeShardPointsReq
 
+The `writeShardPointsReq` statistic is incremented for every point in every internal write request from any remote node.
+
 ###writeShardReq
 
+The `writeShardReq` statistic is incremented every time there is an internal write request from a remote node, regardless of success.
```
mark-rushakoff: Yes.

content/enterprise/v1.0/troubleshooting/internal_statistics.md
```diff
@@ -21,152 +21,231 @@ For metrics shared by both InfluxDB and InfluxEnterprise, please refer to the [I
...
 | :---- | :---- | :---- |
-| writeShardReq | integer | bananas |
-| writeShardReqPoints | integer | bananas |
+| writeShardReq | integer | none |
+| writeShardReqPoints | integer | none |
 
 ###writeShardReq
 
+The `writeShardReq` statistic is incremented the first time the hinted handoff engine handles a write for a remote node.
+Subsequent write requests to this node, destined for the same remote node, do not increment this statistic.
```
mark-rushakoff: I think this particular stat was perhaps not incremented everywhere it was intended to be.

The stat will be set to 0 upon `influxd` restart, regardless of state the last time the process was alive. It is incremented  when the HH "supersystem" is instructed to enqueue a write for the node, and the "subsystem" for the destination node doesn't exist and has to be created, and the "subsystem" created successfully.

The code looks like, if HH files are on disk for a remote node at process startup, the branch that increments this stat will not be reached.

content/enterprise/v1.0/troubleshooting/internal_statistics.md
```diff
@@ -21,152 +21,231 @@ For metrics shared by both InfluxDB and InfluxEnterprise, please refer to the [I
...
+
 ###writeShardReqPoints
 
+The `writeShardReqPoints` statistic is incremented for each point in the first request to the hinted handoff engine for a remote node.
```
mark-rushakoff: This statistic is set in the same place as `writeShardReq` with the same caveats.

I think for 1.1 we can either fully drop the `hh` measurement in favor of the `hh_processor` measurement, or we need to move the code that updates these stats into the hot path, changing their behavior drastically from 1.0 behavior.

content/enterprise/v1.0/troubleshooting/internal_statistics.md
```diff
@@ -19,154 +19,242 @@ a few examples showing good and problematic query responses for some or all of t
...
+| queueDepth | integer | none |
+| writeNodeReq | integer | none |
+| writeNodeReqFail | integer | none |
+| writeNodeReqPoints | integer | none |
+| writeShardReq | integer | none |
+| writeShardReqPoints | integer | none |
 
 ###bytesRead
 
+The `bytesRead` statistic is incremented for every byte read from the HH queue and sent to its destination node.
```
mark-rushakoff: > If the bytes are read and a write attempt is made but fails, is this metric incremented?

Kind of. HH writes occur in concurrent batches as determined by the `retry-concurrency` setting. If an individual write succeeds, the metric is incremented. If _any_ write out of the whole batch fails, the entire batch is considered unsuccessful, and every part of the batch will be retried later. This was not the intended behavior of this stat.

The other situation where `bytesRead` could be larger would be after a restart of the process. Say at startup there were 1000 bytes still enqueued in HH from the previous run of the process. Immediately after a restart, both `bytesRead` and `bytesWritten` are set to zero. Assuming HH is properly depleted, and no future writes require HH, then the stats will read 1000 bytes read and 0 bytes written.

It looks like `bytesRead` being multiply-counted is no longer an issue in 1.1. Also, 1.1 introduces the `queueBytes` stat which should accurately and absolutely track the number of bytes of encoded data waiting to be sent to the remote node.

content/enterprise/v1.0/troubleshooting/internal_statistics.md
```diff
@@ -19,154 +19,242 @@ a few examples showing good and problematic query responses for some or all of t
...
+
+Note that if the data node process is restarted while there is data in the HH queue, `bytesRead` may settle to a number larger than `bytesWritten`.
+
 ###bytesWritten
 
+The `bytesWritten` statistic is incremented for every byte written to the HH queue.
+
+Note that it only tracks bytes written during the lifecycle of the current process.
+Upon process restart, the statistic resets to zero.
```
mark-rushakoff: Yes, both `bytesRead` and `bytesWritten` are set to zero at startup.

content/enterprise/v1.0/troubleshooting/internal_statistics.md
mark-rushakoff: This should be pretty much the same as the `write.pointReq` OSS stat. Incoming writes have to make it through a couple checks before reaching this code (points parse correctly, correct authentication provided, etc.). After that, this statistic should be incremented regardless of source (HTTP, UDP, _internal stats, OpenTSDB plugin, etc.).

content/enterprise/v1.0/troubleshooting/internal_statistics.md
mark-rushakoff: `pointReqRemote` is incremented immediately before the remote write attempt, which only happens if HH doesn't exist for that node.  Then if the write attempt fails, we check again if HH exists, and if so, add the point to HH instead.

It doesn't look like we have a stat or even a pair of stats to track "writes that were input to this node but needed to be written to a remote node, and were successfully written without HH."

content/enterprise/v1.0/troubleshooting/internal_statistics.md
mark-rushakoff: For a single write request that comes in over HTTP or another input method, `writeTimeout` will be incremented by 1 if the entire batch isn't written within the timeout, regardless of whether the points within the batch can be written locally or remotely.

mark-rushakoff:
Got back to this earlier than I expected. I added detailed answers to all the comments, but I didn't make any changes to the markdown at this time.

evbo:
How does this sound?:

```
###PauseTotalNs

The total time in nanoseconds that garbage collection has paused all running Go routine .tick scripts.
```

I took that description from here: https://groups.google.com/forum/#!topic/golang-nuts/hwQ2DwoecK0

Also, does anyone know what these are?:

1. graphite.batchesTx
1. graphite.batchesTxFail
1. graphite.bytesRx
1. subscriber.writeFailures
1. httpd.req
1. queryExecutor.queriesActive

rkuchan:
Closing after a verbal discussion to not document the `_internal` database.

blalor:
What's the story behind not documenting `_internal`?  This understanding seems pretty crucial to understanding, monitoring, and tuning InfluxDB.

seletskiy:
I agree with @blalor, `_internal` should be documented. Without it it's impossible to understand what's going on in your InfluxDB instance.

PR #779, opened 2016-10-07
Fix version numbers, oops

rkuchan:
I messed up the Cluster version numbers - it's 1.0.2-c1.0.3 not 1.0.3-c1.0.3.


rkuchan:
Actually - they're no longer in the s3 bucket. Updating now.

PR #781, opened 2016-10-10
InfluxQL function stddev() is an aggregation function

Tomcat-Engineering:
The standard deviation function is an **aggregation**: it takes in many values and always returns a single value.  It should be categorised with the other aggregation functions like `mean` (to which it is closely related mathematically).

It was previously categorised with the **transformation** functions, which all take a time series and return another time series.


beckettsean:
@desa @toddboom, thoughts?

desa:
I agree with @Tomcat-Engineering

Issue #785, opened 2016-10-11
Code Block formatting is trimming whitespace incorrectly


nathanielc:
All the whitespace on the code blocks is getting left trimmed making it hard to read.

The python code is completely wrong and cannot be understood:

https://docs.influxdata.com/kapacitor/v1.0/examples/anomaly_detection/ 

TICKscripts are hard to read:

https://docs.influxdata.com/kapacitor/v1.0/nodes/join_node/ 

My memory says that the left whitespace was preserved and that this behavior has changed recently.

PR #787, opened 2016-10-11
Fix codeblock formatting

rkuchan:
Fixes https://github.com/influxdata/docs.influxdata.com/issues/785.


kostasb:
LGTM

http://docs.staging.influxdata.com:31004/kapacitor/v1.0/nodes/join_node/

PR #806, opened 2016-10-14
Clarify use of query parameter `rp` for writes

mark-rushakoff:


beckettsean:
thanks, @mark-rushakoff!

PR #809, opened 2016-10-15
Rename node-influx to `influx` on the clients page

connor4312:
The repo is named [node-influx](https://github.com/node-influx/node-influx), primarily for Google-ability and historical reasons I believe (I just joined / took over maintenance this year), but the package itself is published in the JavaScript ecosystem as simply [`influx`](https://www.npmjs.com/package/influx). The package actually called `node-influx` [appears to be](https://www.npmjs.com/package/node-influx) a virtual currency.


beckettsean:
Thanks, @connor4312

Issue #542, opened 2016-07-21
Glossary pages for all products


rkuchan:
Enterprise, Kapacitor, and Telegraf need a Concepts/Glossary page.

rkuchan:
Also see https://github.com/influxdata/docs.influxdata.com/issues/541

beckettsean:
@nathanielc @rossmcdonald what would you put into a Kapacitor glossary that isn't already on the [InfluxDB glossary](http://docs.influxdata.com/influxdb/v1.0/concepts/glossary/)? E.g. `DAG`, `TICKscript node`, etc.

@sparrc @kostasb what would you put into a Telegraf glossary that isn't already on the [InfluxDB glossary](http://docs.influxdata.com/influxdb/v1.0/concepts/glossary/)? E.g. `input plugin`, etc.

kostasb:
I'd say:

`agent`,`input plugin`,`service input plugin`,`output plugin`, `interval`, `flush_interval`, `metric buffer`, `collection jitter`, `flush jitter`

We will need to add aggregator plugins in the future.

sparrc:
yes, also add `aggregator plugin` `processor plugin` `batch size` `precision`

nathanielc:
For Kapacitor:

chaining method
property method
pipeline/DAG
TICKscript nodes/edges
lambda
stream
batch
task
template
replay
recording
time - not as a glossary item but rather how time works inside Kapacitor


That would be a good start.

PR #866, opened 2016-11-08
Create glossary.md

beckettsean:
partial fix for #542

### DIFF
--- a/content/telegraf/v1.1/concepts/glossary.md
+++ b/content/telegraf/v1.1/concepts/glossary.md
@@ -8,43 +8,52 @@ menu:
 
 ## agent
 
-Agent is the core part of Telegraf that gathers metrics from the declared inputs and sends to the declared outputs, based on the plugins enabled by the given configuration.
+An agent is the core part of Telegraf that gathers metrics from the declared inputs and sends to the declared outputs, based on the plugins enabled by the given configuration.
+
+Related entries: [input plugin](/telegraf/v1.1/concepts/glossary/#input-plugin), [output plugin](/telegraf/v1.1/concepts/glossary/#output-plugin)
 
 ## aggregator plugin 
 
-Aggregator Plugins create aggregate metrics and run in between input and output plugins.
+Aggregator Plugins receive raw metrics from input plugins and create aggregate metrics from them. 
+The aggregate metrics are then passed to the configured output plugins.
 
-Related entries: 
+Related entries: [input plugin](/telegraf/v1.1/concepts/glossary/#input-plugin), [output plugin](/telegraf/v1.1/concepts/glossary/#output-plugin)
 
 ## batch size 
 
-The Telegraf agent will send metrics to outputs in batches. The batch size controls the size of writes that Telegraf sends to output plugins.
+The Telegraf agent sends metrics to outputs in batches, not individually. 
+The batch size controls the size of each write batch that Telegraf sends to the output plugins.
+
+Related entries: [output plugin](/telegraf/v1.1/concepts/glossary/#output-plugin)
 
 ## collection interval
 
-The global interval configured for the agent to collect data from all input plugins. The collection interval can be overridden by each individual plugin's configuration.
+The default global interval for collecting data from each input plugin. 
+The collection interval can be overridden by each individual input plugin's configuration.
 
-Related entries: 
+Related entries: [input plugin](/telegraf/v1.1/concepts/glossary/#input-plugin)
 
 ## collection jitter
 
-Collection jitter is used to introduce jitter to the metrics collection by a random amount.
-Each plugin will sleep for a random time within jitter before collecting.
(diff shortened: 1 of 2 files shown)


beckettsean:
@sparrc @kostasb the glossary page is ready for some details. Please add whatever you can, don't worry about the language. Regan and I can clean it up.

content/telegraf/v1.1/concepts/glossary.md
```diff
@@ -0,0 +1,97 @@
...
+
+Related entries: [output plugin](/telegraf/v1.1/concepts/glossary/#output-plugin)
+
+## output plugin
+
+Output plugins deliver metrics to their configured destination. In order to activate an output plugin, it needs to be enabled and configured in the corresponding section in the given config file.
+
+Related entries: [aggregator plugin](/telegraf/v1.1/concepts/glossary/#aggregator-plugin), [flush interval](/telegraf/v1.1/concepts/glossary/#flush-interval), [input plugin](/telegraf/v1.1/concepts/glossary/#input-plugin), [processor plugin](/telegraf/v1.1/concepts/glossary/#processor-plugin)
+
+## precision
```
beckettsean: @kostasb does this control the precision of the emitted metrics? Would it change the timestamps passed to aggregator or processor plugins?

content/telegraf/v1.1/concepts/glossary.md
```diff
@@ -0,0 +1,97 @@
...
+## processor plugin 
+
+Processor Plugins transform, decorate, and/or filter metrics and run in between input and output plugins.
+
+Related entries: [aggregator plugin](/telegraf/v1.1/concepts/glossary/#aggregator-plugin), [input plugin](/telegraf/v1.1/concepts/glossary/#input-plugin), [output plugin](/telegraf/v1.1/concepts/glossary/#output-plugin)
+
+## service input plugin
+
+Service input plugins are input plugins that run in a passive collection mode while the Telegraf agent is running. 
+Service input plugins listen on a socket for known protocol inputs, or actively apply their own logic to collect metrics and deliver them to the Telegraf agent at will.
```
beckettsean: So are service input plugins passive or not? Can you point me to examples of ones that are active? Why aren't they just input plugins?

content/telegraf/v1.1/concepts/glossary.md
kostasb: I got this from some code comments, however after looking at the code it seems to me that precision is always set to nanoseconds. Modifying this parameter does not seem to affect the points precision. We need feedback from @sparrc for this.

content/telegraf/v1.1/concepts/glossary.md
sparrc: All are passive servers, such tcp_listener, http_listener, udp_listener, kafka_consumer, mqtt_consumer, statsd, etc.

kostasb: They are passive in a sense that they all just sit there and wait for inputs. The word "actively" doesn't fit there, my bad.

content/telegraf/v1.1/concepts/glossary.md
kostasb: So the timestamp's precision is truncated but write precision remains in ns.

E.g. 1479295580998402598 will be truncated to 1479295580 and then filled with trailing 0's to reach ns precision again as 1479295580000000000

kostasb: @sparrc just to clarify, I see that the accumulator sets the precision on per-plugin basis, so input plugins deliver their data through Gather to the agent using the configured precision. 
So it changes the timestamp before points are passed to aggregator or processor plugins.

Is this correct?

sparrc: @kostasb yes, that is correct

beckettsean: @sparrc does this commit seem correct, then? https://github.com/influxdata/docs.influxdata.com/pull/866/commits/2fcfb9fff87917e6ac2e6b09e48a4eb33ac29ecb

kostasb:
After quite some digging in to the indexing order, I think this is ready to get a review & merge @rkuchan !

rkuchan:
@kostasb - weights look good :)

I made a couple small changes. If you approve of them I think this is good to go. I did change `output plugins must emit all timestamps` to `input plugins must emit all timestamps` in the definition of precision. I think that's right but feel free to change it back if my understanding is wrong.

kostasb:
@rkuchan thanks for pointing this out, I guess it was unclearly written
output plugins only support nanoseconds precision. input plugins use the configurable precision setting to truncate timestamps to the desired precision, but telegraf will then pad the timestamp with zeros to bring it back to ns precision 

Please feel free to modify the precision section if you feel that the wording doesn't feel right

rkuchan:
@kostasb - got it. Thanks!

PR #872, opened 2016-11-11
Create before_you_begin

beckettsean:
@rkuchan @jwilder @kostasb @jackzampolin @rossmcdonald 

First pass at a "Cluster stuff you can't ignore!" doc. Lemme know what you think. Would like to merge before Monday.

### DIFF
--- a/content/enterprise/v1.1/introduction/installation_guidelines.md
+++ b/content/enterprise/v1.1/introduction/installation_guidelines.md
@@ -1,5 +1,5 @@
 ---
-title: Installation Guidelines
+title: Installation Guidelines (⏰ Please Read!)
 aliases:
   /enterprise/v1.1/introduction/meta_node_installation/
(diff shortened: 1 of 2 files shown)


jackzampolin:
I feel like you should rename `Use three and only three Meta nodes` to `USE THREE AND ONLY THREE META NODES!!!` but that might be too aggressive. Looks great @beckettsean

rkuchan:
@beckettsean, I was thinking that this could live in `Installation Guidelines` - what do you think? I could change the title to `Installation Guidelines (Please Read)`.

I like that structure because it's consistent with the format of the docs for the other projects. I am, however, willing to break from that consistency if you feel strongly about it.

beckettsean:
@rkuchan that works well for me, and I defer to your judgment.

beckettsean:
@rkuchan I'll let you edit this branch directly and merge when you are ready, okay?

PR #886, opened 2016-11-16
Fix for build issue

rkuchan:
Adds an item to the deprecated section's sidebar to get rid of the build error below:
```
ERROR: 2016/11/16 template: partials/chronograf/sidebar.html:26:35: executing "partials/chronograf/sidebar.html" at <.Identifier>: range can't iterate over <*hugolib.Menu Value> in partials/chronograf/sidebar.html
```


rkuchan:
@kostasb, this should do it!

kostasb:
@rkuchan Yes, it does!

PR #887, opened 2016-11-17
Create derivative_across_series.md

adrien-mogenet:
Describe a common use-case, IMHO, with a possible use of `k8sAutoscale`.

### DIFF
--- a/content/kapacitor/v1.1/examples/derivative_across_series.md
+++ b/content/kapacitor/v1.1/examples/derivative_across_series.md
@@ -1,0 +1,50 @@
+---
+title: Aggregate derivative across series
+
+menu:
+  kapacitor_1_1:
+    name: Aggregate derivative across series
+    identifier: derivative_across_series
+    weight: 0
+    parent: examples
+---
+
+Some components are exposing monotonically increasing series such as a webserver with the raw number of total input metrics.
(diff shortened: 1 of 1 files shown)


rossmcdonald:
I think this looks great, and we could definitely use more examples in the documentation.

@nathanielc What do you think?

content/kapacitor/v1.1/examples/derivative_across_series.md
```diff
@@ -0,0 +1,50 @@
...
+    |groupBy('cluster_id')
+    |sum('queries_per_sec')
+        .as('queries_per_sec')
+    |window()
+        .period(5m)
+        .every(1m)
+        .align()
+    // Keeping the integer part should be enough
+    |eval(lambda: int("queries_per_sec"))
+        .as('queries_per_sec')
```
nathanielc: I'd like to see a final use of the resultant calculation. A simple solution would be to use an `influxDBOut()` node and write the data back to a temporary database measurement. Or as you suggest show how it can be used with the `k8sAutoscaler` node. 

In either case I would like to see the example script to do something a user can observe without needing to modify the script.

rkuchan:
We really appreciate the contribution, @adrien-mogenet. I'm closing this for now due to inactivity but feel free to re-open if you'd like to make the requested changes!

PR #888, opened 2016-11-17
Update README

rkuchan:
Adds some content to the README to encourage contribution and give people an idea of what we're looking for. We don't have a lot of style requirements at the moment but I listed a couple items so that there's a little guidance. 

I also added a description for InfluxEnterprise on the docs landing page (see below) and updated the CONTRIBUTING guide to say InfluxData instead of InfluxDB.
image


beckettsean:
👍

Issue #808, opened 2016-10-15
Incorrect description of syntax in example authentication using HTTP API


hsundararaja:
###### URL for relevant page?

https://docs.influxdata.com/influxdb/v1.0/query_language/authentication_and_authorization/#authenticate-using-the-http-api
###### What products and version are you using?

influxDB 
###### Where did you look before opening the issue?

https://docs.influxdata.com/influxdb/v1.0/query_language/authentication_and_authorization/#authenticate-using-the-http-api

There's a message saying -

Authenticate by providing query parameters in the URL. Set u as the username and p as the password.

`curl -G http://localhost:8086/query --data-urlencode "u=todd" --data-urlencode "p=influxdb4ever" --data-urlencode "q=SHOW DATABASES"`

If you do --data-urlencode, it becomes part of the request body. If it should be part of the request parameters, it should be mentioned the following way -

`curl -G 'http://localhost:8086/query?u=todd&p=influxdb4ever' --data-urlencode "q=SHOW DATABASES"
`

beckettsean:
@hsundararaja the credentials are not Basic Auth, they are being passed to the database. It works perfectly well for them to be part of the request body.


```
curl -G http://<redacted>:8086/query --data-urlencode "u=åß∂ƒ" --data-urlencode "p=<redacted>" --data-urlencode "q=SHOW DATABASES"
{"results":[{"series":[{"name":"databases","columns":["name"],"values":[["mydb"],["_internal"],["foo"]]}]}]}
```

hsundararaja:
I know, I didn't say they don't work. I just meant the documentation is wrong when it says "query parameters in the URL". 

Query parameters - 
http://localhost:8086/query?u=todd&p=influxdb4ever 

Body of request - 
--data-urlencode "u=todd" ---data-urlencode "p=influxdb4ever"

The document says "query parameters", but the example is as "body of request". 

Both work though,.

beckettsean:
Thanks for the clarification, @hsundararaja. The curl command is not wrong, just the text describing it.

hsundararaja:
Oops yeah,sorry, thanks for changing the title. That's what I meant.

Issue #807, opened 2016-10-14
add subscription to glossary


beckettsean:

beckettsean:
fixed by https://github.com/influxdata/docs.influxdata.com/pull/891

PR #891, opened 2016-11-18
multiple minor fixes

### LINKED ISSUE #821: Continuous Query example range clarification
###### URL for relevant page?

https://docs.influxdata.com/influxdb/v1.0/query_language/continuous_queries/

Influxdb

The rest of the documentation
###### What is the issue?

The following statement in the continuous querys docs is inaccurate:

> When the CQ executes, it runs a single query for the time range between now() and now() minus the GROUP BY time() interval. If the GROUP BY time() interval is one hour and the current time is 17:00, the query’s time range is between 16:00 and 16:59.

This states that the last minutes worth of data for each hour would never be included, which is not the case.

Instead something like the following would be better:

> When the CQ executes, it runs a single query for the time range between now() and now() minus the GROUP BY time() interval. If the GROUP BY time() interval is one hour and the current time is 17:00, the query’s time range is >= 16:00 and < 17:00.

beckettsean:
fixes #821, #807, #808


beckettsean:
@rkuchan ready for review

PR #893, opened 2016-11-20
Incorrect curl syntax

leitzler:
curl -GET is the same as curl -G -E -T and that's probably not intended here


rkuchan:
Thanks for catching that, @leitzler!

PR #896, opened 2016-11-22
Typo in explanation

takirala:
The time stamp is incorrect in the explanation. Updated it to appropriate time stamp.


rkuchan:
@takirala, that's actually the intended text. For the Continuous Query in question, InfluxDB automatically groups results into preset hours; those hours always start at the top of the hour and end at the bottom of the hour. So even though InfluxDB runs the query at 8:30, it still queries the data for the relevant preset hour: between 8:00 and just before 9:00.

We're always looking to make the documentation clearer so we'll keep this section on the CQ page in mind as we move forward. Thank you for the contribution and I hope this helps!

PR #901, opened 2016-11-26
Fixing grammar.

matsola:
Seems to me (even though english is not my mother tongue) that this sentence is wrong.

Best regards,
Influx noob


rkuchan:
You're totally right, @matsola. Thank you for catching this and for fixing it!

PR #904, opened 2016-11-29
move repo READMEs to appropriate places

### LINKED ISSUE #769: add some repo READMEs to telegraf docs
https://github.com/influxdata/telegraf/blob/master/docs/ has a lot of good detail that should probably be added to the docs site. We also should include a link back to that directory for any new, experimental, or fast-moving READMEs that we don't yet want to add to the docs site.

beckettsean:
fixes #769


beckettsean:
requires https://github.com/influxdata/docs.influxdata.com/pull/866 to be complete, as it assumes the presence of the Glossary in the Concepts section.

PR #914, opened 2016-12-07
Remove insert into note

corylanou:
This note was due to an incorrect behavior in the client.  The bug has been fixed in the client, so the note no longer applies.

//cc @rkuchan


rkuchan:
Fixes: https://github.com/influxdata/docs.influxdata.com/issues/952

rkuchan:
Ported over your changes to the 1.2 docs in #945. 😃

PR #916, opened 2016-12-13
Wrong field name in alternate example

matsola:
Shouldn't it be "usage_idle" as in the above example.


rkuchan:
@matsola, thank you for catching that! Really appreciate it!

PR #930, opened 2017-01-05
Fix typo: downsample_orders -> downsampled_orders

lf-:


rkuchan:
@lf-, good catch! Thank you for fixing.

PR #931, opened 2017-01-05
Add no data FAQ

rkuchan:
Adds a new entry to the FAQ page: Why don't my queries return any data?

Taken from https://www.influxdata.com/tldr-influxdb-tech-tips-december-22-2016/.

### DIFF
--- a/content/influxdb/v1.1/troubleshooting/frequently-asked-questions.md
+++ b/content/influxdb/v1.1/troubleshooting/frequently-asked-questions.md
@@ -43,6 +43,7 @@ Where applicable, it links to outstanding issues on GitHub.
 **Querying data**  
 
 * [What determines the time intervals returned by `GROUP BY time()` queries?](#what-determines-the-time-intervals-returned-by-group-by-time-queries)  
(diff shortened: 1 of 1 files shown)


rkuchan:
@mark-rushakoff - we discussed adding an FAQ about this a couple weeks ago. Would you mind giving this a quick read? I tried to cover the most common explanations but I can always add more!

rkuchan:
Re-worded after a discussion with @mark-rushakoff.  Is this better?

content/influxdb/v1.1/troubleshooting/frequently-asked-questions.md
```diff
@@ -512,6 +513,24 @@ time                    sunflowers                 time                  mean
                        |--|
 ```
 
+## Why don't my queries return any data?
```
mark-rushakoff: Can you change the heading to something that covers not only the "no data" case but also includes "partial data" or "missing some data"? I'm a little concerned that people in the latter case may overlook this FAQ because they would have partial results.

LGTM after that.

rkuchan:
@mark-rushakoff, thank you!

PR #934, opened 2017-01-06
Clarify PagerDuty integration instructions

jeremyd:
This confused me when setting up PagerDuty I accidentally created a Service API key.  Not an integration key as is required.  Hoping this helps others!


rossmcdonald:
These changes look good to me. @nathanielc What do you think?

rossmcdonald:
Thanks @jeremyd!

PR #935, opened 2017-01-06
Fix typo for schema numbering

bayendor:
Should be schema 1 for first example "Don't put more than one piece of information in one tag"


rkuchan:
Thank you, @bayendor!

Issue #933, opened 2017-01-05
[InfluxDB] Regular expressions in the GROUP BY clause


rkuchan:
Data Exploration

InfluxQL supports using regular expressions to specify tag keys in the `GROUP BY` clause:
```
> SELECT FIRST("index") FROM "h2o_quality" GROUP BY /l/

name: h2o_quality
tags: location=coyote_creek
time                  first
----                  -----
2015-08-18T00:00:00Z  41

name: h2o_quality
tags: location=santa_monica
time                  first
----                  -----
2015-08-18T00:00:00Z  99
```

Will need to update the `GROUP BY` tags section as well as the Regular Expressions in Queries section.

Issue #932, opened 2017-01-05
Security best practices


rkuchan:
Best practices, and this will require a separate HTTPS setup guide as well.

Issue #927, opened 2016-12-29
[InfluxDB] Update config doc


rkuchan:
- [x] The section headers are no longer commented out in the default configuration file, https://github.com/influxdata/influxdb/pull/7681 - will also need to update the FAQ for this one, and the admin page.
- [x] Collectd has new settings: `security-level` and `auth-file` (https://github.com/influxdata/influxdb/pull/7435/files#diff-738d25ddfcbd646924a0fb88654fc48bR343)

And do a general run-through.

Issue #924, opened 2016-12-21
[InfluxDB] Documentation on default ports no longer list 8088


hraftery:
The Installation documentation page lists 8083 and 8086 as the default ports in use by InfluxDB. The default influxdb.conf backs this up (though also suggests 8089 for UDP line protocol data).

However, the backup_and_restore documentation page uses port 8088 in its example for getting a backup from a remote host. This works just fine on a default install.

https://github.com/influxdata/influxdb/issues/6286 reveals that the port used to be explicit, but was removed because the open source version does not support clustering. However the use of the port remains, as does the backup functionality.

It seems less confusion would result from including 8088 in the list of default ports, and perhaps mentioning that its use in the open source product is restricted to remote backups (if that's true).

###### URL for relevant page?
https://docs.influxdata.com/influxdb/v1.1/introduction/installation/

###### What products and version are you using?
InfluxDB 1.1.1

###### Where did you look before opening the issue?
https://docs.influxdata.com/influxdb/v1.1/introduction/installation/
https://docs.influxdata.com/influxdb/v1.1/administration/backup_and_restore/
/etc/influxdb/influxdb.conf
https://docs.influxdata.com/influxdb/v1.1/tools/api/
Various Google searches on "influxdb" "port" "8088" and "8086"

rkuchan:
@hraftery, thank you for pointing this out. Our documentation on ports was lacking. We've updated the docs for the 1.2 release with more information. Again, really appreciate the feedback!

PR #945, opened 2017-01-12
[InfluxDB, Enterprise, Telegraf] V1.2 docs

rkuchan:
More to come. 

Creates v1.2 directories for InfluxDB, Enterprise, and Telegraf. Holding off on Kapacitor as those docs tend to come in separately and Chronograf because that comes out on Friday. I make updates for 1.2 in all subsequent commits (commits are done per issue). 

### InfluxDB updates
* #927: Updates docs for any configuration changes. Updates the config doc, the faq doc, and the web admin doc - config section headers are now uncommented in the default config. Also adds the new Collectd config settings and port `8088` to the config doc and moves some things around in that doc to match the organization of the default config.
* #924: Updates the port documentation to include `8088`. This touches the ports, installation, and backup and restore pages. No longer includes localhost in the config bind address - that's moved to 1.3.
* #883: Removes a now unnecessary common issue with the `SAMPLE()` function.
* #909: Updates the configuration doc and the querying data with the HTTP API guide with `max-row-limit`'s new `"partial":true` tag. This also adds a note to the API reference about `max-row-limit` because that was missing.
* #933: Adds `GROUP BY /<regex>/` to the data exploration and influxql spec pages. This functionality isn't specific to 1.2, it was just missing from the docs altogether. 
* #951: Adds documentation for subqueries to the data exploration page and link to that section from the CQ doc and the FAQ doc.
* #953: Updates the database management page to include the minimum allowable `SHARD GROUP DURATION`. Thanks to Cory for submitting a PR on this.
* #949, #950: Adds documentation for the `clear` and `use database.retention_policy` commands in the CLI. Also adds retention policy to the list of items returned by the `settings` command. Thanks to Cory for submitting a PR on this.
* #952: Removes the note about the retention policy persisting after a single `INSERT INTO` command in the CLI. Also thanks to Cory for submitting a PR on this.
* #856: Updates the `CREATE DATABASE WITH` documentation to reflect that `autogen` doesn't get created if you create a retention policy too.
* #948: Updates the differences pages and the upgrading page.
* Removes outdated mentions of version 1.1.
* #932: Adds a security best practices guide and a guide for setting up HTTPS with InfluxDB.

### Enterprise updates
* #961: Removes the need to uncomment the `[http]` section header during the quickstart and production installation processes. Config section headers are now uncommented in the default config.
* #959: Updates the clustering version and web console version to from `1.1.1` to `1.2.0`.
* #962: Updates the Changelog/Release notes page with the information in https://github.com/influxdata/plutonium/blob/master/CHANGELOG.md. Attempts to reword some of the issue titles to be more applicable.
* #960: Updates the backup docs - backups are now incremental by default and there's a new `-full` flag.
* #958: Updates the upgrading doc - this may need more editing depending on the web console installation process. This also updates some of the installation steps - nothing major, just small updates.
* Removes outdated mentions of version 1.1.
* #966: Adds a manual rebalance doc and updates the changelog and the web console features page with the deprecation of the rebalance button.

### Telegraf updates
* #957: Updates the content of any Telegraf documentation that comes directly from the repo. In this case, just `data_formats_input.md` required some changes. This also removes a page that shouldn't have been in the docs.
* #955: Updates the plugins lists: 1.2 just has the new `internal` input and `discard` output plugins.
* #956: Updates the differences and upgrading pages.


content/enterprise/v1.2/guides/rebalance.md
```diff
@@ -0,0 +1,204 @@
...
+
+This guide describes how to manually rebalance an InfluxEnterprise cluster.
+Rebalancing a cluster involves two primary goals:
+
+* Evenly distribute
+[shards](/influxdb/v1.2/concepts/glossary/#shard) across all data
+nodes in the cluster and second
+* Ensure that every
+[shard group](/influxdb/v1.2/concepts/glossary/#shard-group) contains N number
+of shards, where N is determined by the retention policy's
```
pauldix: this is off. The replication factor sets the number of `owners` you'd find for a shard. The number of shards within a shard group is generally determined by `# of nodes / rf` so like `4 / 2`. This behavior is different if your node count isn't evenly divisible by your RF. In that case it'll find the lowest number of shards it would have to have in the group to have an equal number of shards on each node.

content/enterprise/v1.2/guides/rebalance.md
```diff
@@ -0,0 +1,204 @@
+---
+title: Cluster Rebalance
+menu:
+  enterprise_1_2:
+    weight: 19
+    parent: Guides
+---
+
+This guide describes how to manually rebalance an InfluxEnterprise cluster.
```
pauldix: It might be worth mentioning that features for automatically rebalancing a cluster based on different criteria are currently under development.

content/enterprise/v1.2/guides/rebalance.md
```diff
@@ -0,0 +1,204 @@
...
+The steps below are applicable for any cluster size.
+
+By the end of this section you will have a fully balanced cluster with
+the relevant historical shards moved from one of the old data nodes to the
+new data node.
+
+#### Step 1: Truncate Hot Shards
+
+Hot shards are shards that are currently receiving writes.
+Performing any action on a hot shard can lead to data loss.
```
pauldix: This isn't totally accurate. It could lead to data inconsistency within the cluster, which would have to be manually recovered later.

content/enterprise/v1.2/guides/rebalance.md
```diff
@@ -0,0 +1,204 @@
...
+
+The command below truncates all hot shards; the system creates a new
+hot shard which is automatically distributed across all data nodes in the cluster.
+InfluxDB writes all new points to that new shard.
+All previous writes are now stored in cold shards.
```
pauldix: it's worth mentioning that while you're performing any of these actions, they should refrain from doing historical data loads. Basically, if you're doing this kind of cluster maintenance, hold off on historical loads.

content/enterprise/v1.2/guides/rebalance.md
```diff
@@ -0,0 +1,204 @@
...
+> **Note** Find the TCP address of the new data node with the `influxd-ctl show`
+command.
+
+#### Step 4: Remove Unnecessary Cold Shards
+
+Once the relevant cold shards are copied to the new data node, remove
+the shard from the old data node with the following command:
+
+```
+remove-shard <source_TCP_address> <shard_ID>
```
pauldix: you should make it very clear that this is an irrecoverable destructive action. Measure 5 times and cut once.

content/enterprise/v1.2/guides/rebalance.md
```diff
@@ -0,0 +1,204 @@
...
+```
+
+Where `source_TCP_address` is the address that you noted in step 2.
+
+#### Step 5: Confirm the Rebalance
+
+Check the `Owners` column in the output from the following command.
```
pauldix: Confirmation of owners should happen twice in this flow. Once before they remove the shard and once after they remove it. Further, as a safe measure, they should check the shard directory on each of the new owners and very if that it matches up with the old owner. Verification through this method is the most reliable way to be sure that the data has been copied successfully.

content/enterprise/v1.2/guides/rebalance.md
```diff
@@ -0,0 +1,204 @@
...
+The steps below are applicable for any cluster size.
+
+By the end of this section you will have a fully balanced cluster with:
+
+* all historical shards copied and distributed to the new data node
+* all new shards automatically appearing on all data nodes in the cluster
+
+#### Step 1: Update the Retention Policy
+
+[Update](/influxdb/v1.2/query_language/database_management/#modify-retention-policies-with-alter-retention-policy)
```
pauldix: mention this is optional. They may want to keep the RF the same and just expand the disk size or write throughput of the cluster.

content/enterprise/v1.2/guides/rebalance.md
```diff
@@ -0,0 +1,204 @@
...
+```
+ALTER RETENTION POLICY <retention_policy_name> ON <database_name> REPLICATION 3
+```
+
+#### Step 2: Truncate Hot Shards
+
+Hot shards are shards that are currently receiving writes.
+Performing any action on a hot shard can lead to data loss.
+
+The command below truncates all hot shards; the system creates a new
```
pauldix: The assumption is that the only hot shards are ones that are in the current window of time. If they're writing historical data or have lagged data collection, this changes things. They'd need to wait until the historical write is done. And do the truncate and then wait until the period of lagged collection has passed.

pauldix: Just realized after reading further that this could be in the section of "finding cold shards"

PR #988, opened 2017-01-30
[Enterprise] Update backup restore documentation

rkuchan:
This updates the Enterprise docs with the changes in https://github.com/influxdata/plutonium/pull/826.

My current understanding is:

* Full restores require the `-full` flag and the path to the full backup's manifest file
* Incremental restores require the path to the incremental backup's directory

I'm having trouble with both full restores and incremental restores:

### Full restore

```
$ mkdir my-full-backup

$ influxd-ctl backup -full my-full-backup
Backing up meta data... Done. 481 bytes transferred
Backing up node rk-restore-01:8088, db _internal, rp monitor, shard 1... Done. Backed up in 43.653316ms, 396800 bytes transferred
Backing up node rk-restore-01:8088, db telegraf, rp autogen, shard 2... Done. Backed up in 28.324242ms, 182784 bytes transferred
Backed up to my-full-backup in 73.695458ms, transferred 580065 bytes

$ ls my-full-backup/
20170130T195547Z.manifest
20170130T195547Z.meta
20170130T195547Z.s1.tar.gz
20170130T195547Z.s2.tar.gz

$ influxd-ctl restore -full my-full-backup/20170130T195547Z.manifest
Using manifest: my-full-backup/20170130T195547Z.manifest
Restoring meta data... Error.
restore: operation exited with error: problem setting snapshot: cannot restore into non-empty cluster
```
I think this might be because I've written new data to InfluxDB from when I performed the backup.

### Incremental restore

```
$ mkdir my-incremental-backuproot@rk-restore-01:~# influxd-ctl backup my-incremental-backup/
Backing up meta data... Done. 481 bytes transferred
Backing up node rk-restore-01:8088, db _internal, rp monitor, shard 1... Done. Backed up in 43.462854ms, 467456 bytes transferred
Backing up node rk-restore-01:8088, db telegraf, rp autogen, shard 2... Done. Backed up in 28.220163ms, 192000 bytes transferred
Backed up to my-incremental-backup/ in 73.292524ms, transferred 659937 bytes

$ ls my-incremental-backup/
20170130T200202Z.manifest
20170130T200202Z.meta
20170130T200202Z.s1.tar.gz
20170130T200202Z.s2.tar.gz

$ influxd-ctl restore my-incremental-backup/
Using backup directory: my-incremental-backup/
Using meta backup: 20170130T183945Z.meta
restore: open my-incremental-backup/20170130T183945Z.meta: no such file or directory
```

### DIFF
--- a/content/enterprise/v1.2/guides/backup-and-restore.md
+++ b/content/enterprise/v1.2/guides/backup-and-restore.md
@@ -66,6 +66,7 @@ options (for example: they specify a different database with `-db` or a
 different retention policy with `-rp`) must be stored in different directories.
 If a single backup directory stores several different incremental backups, a
 restore only restores the most recent incremental backup.
+This issue will be fixed in the next point release.
 
 ##### Examples
 <br>
@@ -173,7 +174,7 @@ the new cluster.
 #### Restore from an incremental backup
 Use the syntax below to restore an incremental backup to a new cluster or an existing cluster.
 Note that the existing cluster must contain no data in the affected databases.*
-Performing an restore from an incremental backup requires the path to the incremental backup's directory.
+Performing a restore from an incremental backup requires the path to the incremental backup's directory.
 
 <dt> In version 1.2.1, restoring from an incremental backup requires users to `cd` into
 the backup directory and run `influxd-ctl restore [options] .` from that directory.
@@ -186,7 +187,7 @@ influxd-ctl [-bind <hostname>:8091] restore [options] <path-to-backup-directory>
 
 \* The existing cluster can have data in the `_internal` database, the database
 that the system creates by default.
-The system automatically drops the `_internal` database when it performs an incremental restore.
+The system automatically drops the `_internal` database when it performs a complete restore.
 
 Options:
 
@@ -212,7 +213,7 @@ influxd-ctl [-bind <hostname>:8091] restore [options] -full <path-to-manifest-fi
 \* The existing cluster can have data in the `_internal` database, the database
 that the system creates by default.
 The system automatically drops the `_internal` database when it performs a
-full restore.
+complete restore.
 
 Options:
 
@@ -241,13 +242,8 @@ This issue will be fixed in the next point release.
 
 Output:
 ```
-$ cd cd my-incremental-backup/
+$ cd my-incremental-backup/
 $ influxd-ctl restore .
-Using backup directory: .
-Using meta backup: 20170130T231333Z.meta
-Restoring meta data... Error.
-restore: operation exited with error: problem setting snapshot: cannot restore into non-empty cluster
-<hostname>:~/my-incremental-backup# influxd-ctl restore .
 Using backup directory: .
 Using meta backup: 20170130T231333Z.meta
 Restoring meta data... Done. Restored in 21.373019ms, 1 shards mapped
@@ -279,7 +275,7 @@ cd <path-to-backup-directory>
 influxd-ctl restore -db <src> -newdb <dest> .
 ```
 
-<dt> In version 1.2.1, an incremental restore requires users to `cd` into
+<dt> In version 1.2.1, a restore from an incremental backup requires users to `cd` into
 the backup directory and run `influxd-ctl restore [options] .` from that directory.
 This issue will be fixed in the next point release.
 </dt>
@@ -303,7 +299,7 @@ Your `telegraf` database was mistakenly dropped, but you have a recent backup so
 If [Telegraf](/telegraf/v1.2/) is still running, it will recreate the `telegraf` database shortly after the database is dropped.
 You might try to directly restore your `telegraf` backup just to find that you can't restore:
 
-<dt> In version 1.2.1, an incremental restore requires users to `cd` into
+<dt> In version 1.2.1, a restore from an incremental backup requires users to `cd` into
 the backup directory and run `influxd-ctl restore [options] .` from that directory.
 This issue will be fixed in the next point release.
 </dt>
--- a/content/enterprise/v1.2/guides/backup-and-restore.md
+++ b/content/enterprise/v1.2/guides/backup-and-restore.md
@@ -36,31 +36,34 @@ The filenames reflect the UTC timestamp of when the backup was created, for exam
 * Manifest: `20060102T150405Z.manifest`
 
 A **restore** adds the backed-up data to the cluster.
-By default, a restore attempts to write databases using backed-up data's replication factor.
+By default, a restore is incremental and writes to databases using the backed-up data's replication factor.
 An alternate replication factor can be specified with the `-newrf` flag when restoring a single database.
 
 ## Syntax
 
 ### Backup
 ```
-influxd-ctl [-bind <hostname>:8091] backup [options] <backup-directory>
+influxd-ctl [-bind <hostname>:8091] backup [options] <path-to-backup-directory>
 ```
 Options:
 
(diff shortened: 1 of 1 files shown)


joelegasse:
The error you're seeing for restoring with `-full` is `cannot restore into non-empty cluster`. In order to restore, the target cluster needs to be empty. If you're restoring into the same cluster, this can be achieved by dropping all databases (other than `_internal`, which should be dropped for you upon restore)

There is a bug in the `restore` code that you're hitting in the second scenario. Using the current release, you have to `cd` into the backup directory, then run `influxd-ctl restore <OPTIONS> .`, that will be fixed to handle paths correctly once the fix is merged.

rkuchan:
I updated the PR with how to get around the bug in 1.2. @kostas or @joelegasse could you review this before I merge it?

There are a couple things I'm still not completely comfortable with:

* The old doc said: "An incremental restore requires that the cluster hardware has the same hostnames as during the backup. If the hostnames have changed, you will need to restore the databases one-by-one." Is that still true? I couldn't figure out how to get my cluster into a state that I could test that scenario out.
* The previous comment mentioned that the cluster had to be empty for a `-full` restore. I was able to do a `-full` restore on a non-empty cluster by specifying the database to be restored and renaming it with the `-newdb` option. Is that the intended behavior?

content/enterprise/v1.2/guides/backup-and-restore.md
```diff
@@ -36,31 +36,34 @@ The filenames reflect the UTC timestamp of when the backup was created, for exam
 * Manifest: `20060102T150405Z.manifest`
 
 A **restore** adds the backed-up data to the cluster.
-By default, a restore attempts to write databases using backed-up data's replication factor.
+By default, a restore is incremental and writes to databases using the backed-up data's replication factor.
```
joelegasse: Restores are not incremental, just the backups are. Possibly a better phrasing would be that the `restore` command will default to restoring from an incremental backup directory. If a backup was taken with the `-full` flag, then `influxd-ctl restore -full <MANIFEST>` can be used to restore that backup.

The `-full` refers to the backup, not necessarily a "full restore" of the data. That happens by not limiting the restore with `-db`, `-rp`, or `-shard`.

content/enterprise/v1.2/guides/backup-and-restore.md
```diff
@@ -36,31 +36,34 @@ The filenames reflect the UTC timestamp of when the backup was created, for exam
...
 * `-bind <hostname>:8091`: the hostname and HTTP port of a running meta server (defaults to `localhost:8091`)
 * `-db <string>`: the name of the single database to back up
 * `-from <TCP-address>`: the data node TCP address to prefer when backing up
-* `-full`: peform a full backup
+* `-full`: perform a full backup
 * `-rp <string>`: the name of the single retention policy to back up (must specify `-db` with `-rp`)
 * `-shard <unit>`: the ID of the single shard to back up
 
+Restoring a `-full` backup and restoring an incremental backup require different syntax.
+To prevent issues with restore, keep `-full` backups and incremental backups in separate directories.
```
joelegasse: Currently, there is an issue with incremental backups and limiting what is getting backed up. This line in the docs might not be the right place for it, but we should add a recommendation somewhere that unique directories should be used for separate commands. For example `influxd-ctl backup -db foo` and `influxd-ctl backup -db bar` should be given separate directories. Each will perform incremental backups of the respective databases, but with the current bug, they cannot be kept in the same directory, as `restore` will only pick the most recent backup, and not merge the two databases. Even though they come from the same cluster... Internal issue 1004 is tracking that bug.

For now, though, each set of arguments passed to `influxd-ctl backup` should be given a separate directory.

content/enterprise/v1.2/guides/backup-and-restore.md
```diff
@@ -78,12 +81,34 @@ $ ls
...
+Backed up to backup_dir in 51.388233ms, transferred 333793 bytes
+~# ls backup_dir
+20170130T184058Z.manifest
+20170130T184058Z.meta
+20170130T184058Z.s1.tar.gz
+20170130T184058Z.s2.tar.gz
+```
 
-##### Back up a single database
+##### Example 3: Perform an incremental backup on a single database
```
joelegasse: This is probably a good place to add the comment I mentioned above, that each database should be given a unique directory (for now).

content/enterprise/v1.2/guides/backup-and-restore.md
```diff
@@ -97,9 +122,25 @@ $ ls ./telegrafbackup
 ```
 
 ### Restore
+
+#### Incremental Restore
```
joelegasse: I wouldn't call this an "incremental restore". To me, that would imply that you could restore data that has changed since your last restore.

content/enterprise/v1.2/guides/backup-and-restore.md
```diff
@@ -111,75 +152,149 @@ Options:
 * `-rp <string>`: the name of the single retention policy to restore
 * `-shard <unit>`: the shard ID to restore
 
-#### Examples
+> **Note:**
+An incremental restore requires that the cluster hardware has the same hostnames as during the backup.
```
joelegasse: This is not true. 1.2 allows restoring into any empty cluster (`_internal` only).

rkuchan:
@joelegasse  - alright! Made the changes we talked about. Could you take another look at this?

Again - really appreciate all the feedback!

content/enterprise/v1.2/guides/backup-and-restore.md
```diff
@@ -97,9 +152,68 @@ $ ls ./telegrafbackup
...
+
 ```
-influxd-ctl [-bind <hostname>:8091] restore [options] <path-to-manifest-file>
+influxd-ctl [-bind <hostname>:8091] restore [options] -full <path-to-manifest-file>
 ```
+
+\* The existing cluster can have data in the `_internal` database, the database
+that the system creates by default.
+The system automatically drops the `_internal` database when it performs a
+full restore.
```
joelegasse: Earlier you used the phrase "complete backup", I think "complete restore" would fit better here, to prevent confusion with `-full`

content/enterprise/v1.2/guides/backup-and-restore.md
```diff
@@ -113,73 +227,114 @@ Options:
...
 
-`influxd-ctl restore <path-to-manifest>`
+Output:
+```
+$ cd cd my-incremental-backup/
+$ influxd-ctl restore .
+Using backup directory: .
+Using meta backup: 20170130T231333Z.meta
+Restoring meta data... Error.
+restore: operation exited with error: problem setting snapshot: cannot restore into non-empty cluster
```
joelegasse: Is this supposed to show the error message? If so, maybe mention that all databases other than `_internal` were dropped between running the two commands.

content/enterprise/v1.2/guides/backup-and-restore.md
```diff
@@ -97,9 +152,68 @@ $ ls ./telegrafbackup
...
+This issue will be fixed in the next point release.
+</dt>
+
+```
+influxd-ctl [-bind <hostname>:8091] restore [options] <path-to-backup-directory>
+```
+
+\* The existing cluster can have data in the `_internal` database, the database
+that the system creates by default.
+The system automatically drops the `_internal` database when it performs an incremental restore.
```
rkuchan: Darnit! I missed that one. Thank you.

PR #1015, opened 2017-02-13
remove webconsole rebalance reference

kostasb:
webconsole rebalance is deprecated with 1.2

@rkuchan @rossmcdonald @gunnaraasen


rkuchan:
Thank you, @kostasb. I missed that reference to the rebalance button.

PR #1031, opened 2017-02-23
FIx documentation of microseconds duration

cm6051:
The documentation says that `u` and `µ` are supported suffixes for durations, but only `us` and `µs` actually work.

Using `u` or `µ` fails with `run: parse config: time: unknown unit µ in duration 1µ` and `run: parse config: time: unknown unit u in duration 1u`. 

This PR fixes the documentation.

### DIFF
--- a/content/influxdb/v1.2/query_language/spec.md
+++ b/content/influxdb/v1.2/query_language/spec.md
@@ -180,20 +180,20 @@ Durations can be specified with mixed units.
 
 #### Duration units
 
- Units  | Meaning
---------|-----------------------------------------
- u or µ | microseconds (1 millionth of a second)
- ms     | milliseconds (1 thousandth of a second)
- s      | second
- m      | minute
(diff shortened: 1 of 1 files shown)


rkuchan:
Thank you for finding and fixing that, @cm6051. 

I looked into it and this seems to be part of a bigger issue with InfluxDB; the proper microsecond syntax differs for writes, queries, specifying the timestamp `precision` in the CLI, and setting durations in the config file. I've opened an issue on the InfluxDB repo about this: https://github.com/influxdata/influxdb/issues/8053.

For now, I'm not going to merge your docs PR just so the team can figure out what the proper syntax actually is. I've opened a [docs issue](https://github.com/influxdata/docs.influxdata.com/issues/1033) to add a note about the current microsecond syntax behaviour. Thank you again for finding that - we really, really appreciate you letting us know and taking the time to open a docs PR!

PR #1035, opened 2017-02-24
Update hinted handoff - known issues

kostasb:
"The hinted handoff queue becomes blocked if it experiences field type errors."

fixed in enterprise 1.2 with PR #866 thanks to @desa 

It now drops writes that hit field type conflict or max-values-per-tag errors.

@rkuchan


rkuchan:
Closing for now :)

PR #1048, opened 2017-03-07
add where-can-I-find-the-logs section

kostasb:
This question comes up pretty often.

Added systemV and systemd documentation.


kostasb:
just noticed that we have it under http://docs.influxdata.com/influxdb/v1.2/administration/logs/ but.. anyway

rossmcdonald:
Even though it's mentioned the "Logs" section, we also get questions about this all the time, so I'm okay with having it in two places.

PR #1068, opened 2017-03-21
Fix link to old BatchNode doc

austinbutler:


rkuchan:
Thank you, @austinbutler! We appreciate the fix 😃

PR #1070, opened 2017-03-21
[Enterprise] Add known issue about queries with math on >1 selector function

rkuchan:
Adds a known issue about the change in behavior for queries that use math on several selector functions in 1.2.0-1.2.2. It includes a description of the subquery workaround.


rkuchan:
Original issue: https://github.com/influxdata/influxdb/issues/8167

PR #1073, opened 2017-03-23
Updates

pmyjavec:
* Notify the admin user should be created first.
* Add a note about the syntax required when creating new users.

### DIFF
--- a/content/influxdb/v1.2/query_language/authentication_and_authorization.md
+++ b/content/influxdb/v1.2/query_language/authentication_and_authorization.md
@@ -212,7 +212,12 @@ Non-admin users can have one of the following three privileges per database:
 ### User Management Commands
 
 #### Admin user management
-##### `CREATE` a new admin user:  
+
+Before you being able manage the database when HTTP authentication is enabled, you will first need to create the `admin` user:
+
+`CREATE USER admin WITH PASSWORD '<password>' WITH ALL PRIVILEGES`
+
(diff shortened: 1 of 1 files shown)


content/influxdb/v1.2/query_language/authentication_and_authorization.md
```diff
@@ -212,7 +212,12 @@ Non-admin users can have one of the following three privileges per database:
 ### User Management Commands
 
 #### Admin user management
-##### `CREATE` a new admin user:  
+
+Before being able manage the database when HTTP authentication is enabled, you will first need to create the `admin` user:
+
+`CREATE USER 'admin' WITH PASSWORD '<password>' WITH ALL PRIVILEGES`
```
rkuchan: I don't think `admin` should be single quoted; single quoting the username yields a parsing error.

content/influxdb/v1.2/query_language/authentication_and_authorization.md
```diff
@@ -297,6 +302,7 @@ CLI example:
 
 > **Notes:**
 >
+* The user value must be wrapped in double quotes.
```
rkuchan: I'd clarify this a bit; The user value must be wrapped in double quotes if it starts with a digit, contains characters other than `[A-z,0-9,_]`, or if it is an [InfluxQL keyword](https://github.com/influxdata/influxdb/blob/master/influxql/README.md#keywords).

PR #1084, opened 2017-03-31
Add key file permissions example

dstarmanhr:
Key file should be only readable/writable by root.

### DIFF
--- a/content/influxdb/v1.2/administration/https_setup.md
+++ b/content/influxdb/v1.2/administration/https_setup.md
@@ -65,7 +65,7 @@ commands:
 ```
 sudo chown root:root /etc/ssl/<CA-certificate-file>
 sudo chmod 644 /etc/ssl/<CA-certificate-file>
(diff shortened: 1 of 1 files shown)


content/influxdb/v1.2/administration/https_setup.md
```diff
@@ -65,6 +65,7 @@ commands:
 ```
 sudo chown root:root /etc/ssl/<CA-certificate-file>
 sudo chmod 644 /etc/ssl/<CA-certificate-file>
+sudo chmod 600 /etc/ssl/<private-key-file>.key>
```
rkuchan: Could you remove the extra `>` after `<private-key-file`?

dstarmanhr:
I've actually removed the ".key>" so there is consistent naming of files (other don't have extensions).

rkuchan:
👍  Thanks again, @dstarmanhr!

PR #1090, opened 2017-04-04
[InfluxDB] Upgrading, add step for updating permission on meta.db

rkuchan:


kostasb:
Thanks a lot for taking care of this @rkuchan !

PR #1094, opened 2017-04-07
Updates

pmyjavec:
* Notify the admin user should be created first.
* Add a note about the syntax required when creating new users.
* Add some more examples for adding non-root users


rkuchan:
Thank you, @pmyjavec!

PR #1098, opened 2017-04-11
InfluxDB: Queries return series, not tables

jomo:
Please correct me if I'm wrong.


rkuchan:
Hi @jomo! You're correct, queries do return series - the table wording was to describe the tabular output. If that was confusing to you I'd be happy to change it.

jomo:
I found it confusing when I was reading up on the different data types and terms used in InfluxDB. The term `series` is used everywhere else and then suddenly the docs are speaking of `tables` which I thought was a term not used for Influx. I ended up trying to figure out what tables are and how they are different from series.

rkuchan:
Thank you! I'll go ahead and merge this. Please do let us know if there are other confusing word choices - we really appreciate the feedback and your contribution!

PR #1126, opened 2017-05-14
Move "next options control batching" to the correct place

nicolas17:
In the documentation for collectd options, the paragraph saying "The next three options control how batching works" wasn't actually above the three options for batching.

f2cece9a08 added documentation for `retention-policy` between the "batching options follow" paragraph and the `batch-size` option. e80a4b1215 further worsened this by adding `typesdb`, `security-level`, and `auth-file` below `retention-policy`.

This commit moves the paragraph to the correct place, right above `batch-size`.


nicolas17:
Just submitted CLA form.

rkuchan:
Thank you for catching and fixing this, @nicolas17.

PR #1140, opened 2017-05-24
Update the AWS recommended machine class info.

timhallinflux:
We were suggesting C3 class machines --- these are pretty old at this point.

The R4 class machines provide more memory -- which is typically good for the database.


rkuchan:
Looks good! Thank you, @timhallinflux

PR #1142, opened 2017-05-24
examples were switched to guides

timhallinflux:
current link results in a 404. -- as reported by a customer via InfluxData Support.

updated link to point to:
https://docs.influxdata.com/kapacitor/v1.3/guides/using_alert_topics/


rkuchan:
Thank you, @timhallinflux! I put in redirects for the moved pages but I guess they didn't work for that user. I went through and updated some other links in case other users experience that issue.

PR #1143, opened 2017-05-25
Fix wrong `getting-started` link

katopz:


rkuchan:
Thank you for catching and fixing that, @katopz!

PR #1144, opened 2017-05-26
Example for INTO requested by customer

timhallinflux:
Create an example for "rename" database -- which essentially means doing a select into...and keeping tag keys in place.


rkuchan:
Awesome, @timhallinflux! I made some edits - if you're OK with them I'll go ahead and get this merged!

rkuchan:
Got it! I moved it back to the first example.

PR #1165, opened 2017-06-16
Add hyphen example to double-quote section of FAQ

timhallinflux:
Requested change from someone who was struggling with a field/tag coming from the Telegraf sysstats plugin which contained a hyphen.


rkuchan:
Thanks, @timhallinflux!

PR #1181, opened 2017-06-23
Added clarification to Telegraf environment variables

Joe-Influx:
Added clarification to Telegraf environment variables for sensitive data


rkuchan:
Thanks, @Joe-Influx!

PR #1185, opened 2017-06-23
Anti-entropy guide

mark-rushakoff:


content/enterprise_influxdb/v1.3/guides/anti-entropy.md
```diff
@@ -0,0 +1,138 @@
...
+
+The anti-entropy service can only address missing shards when there is at least one copy of the shard available.
+In other words, so long as new and healthy nodes are introduced, a replication factor of 2 can recover from one missing node, a replication factor of 3 can recover from two missing nodes, and so on.
+A replication factor of 1 cannot be recovered by the anti-entropy service if the shard goes missing.
+
+## Scenarios
+
+This section covers some of the common use cases for the anti-entropy service.
+
+### Scenario 1: Replacing a data node that is permanently offline
```
e-dard: I go into more detail in the scenario document.

In short, when replacing a data node in a cluster, I don't think adding a node first, then removing the downed node is the best approach. 

One reason is because when you add the new node it gets a new node ID, which means it's not automatically the owner of the same set of shards that the downed node owned. When you then forcefully remove the downed node the shards will be reassigned to new owners, which will include the newly added node, and as expected the AE service will kick in and pull the shards.

One thing that may confuse operators is that the shards that were on the downed node are not guaranteed to be reassigned to the new node in this case. They actually get assigned to a node with the fewest shards, which in some cases may be the new node, but not always.

The final reason why I don't think this approach is best is that if a cluster is at its entitlement level, it won't be possible to add the new node before removing the old one.

The preferred approach in my opinion for replacing a failed node is to use `update-data`.

content/enterprise_influxdb/v1.3/guides/anti-entropy.md
```diff
@@ -0,0 +1,138 @@
...
+1   _internal  monitor           2                 1            2017-06-22T00:00:00Z  2017-06-23T00:00:00Z  2017-06-30T00:00:00Z  [{5 stable.example:8088}]
+2   _internal  monitor           2                 1            2017-06-22T00:00:00Z  2017-06-23T00:00:00Z  2017-06-30T00:00:00Z  [{6 new.example:8088}]
+4   _internal  monitor           2                 3            2017-06-23T00:00:00Z  2017-06-24T00:00:00Z  2017-07-01T00:00:00Z  [{5 stable.example:8088}]
+5   _internal  monitor           2                 3            2017-06-23T00:00:00Z  2017-06-24T00:00:00Z  2017-07-01T00:00:00Z  [{6 new.example:8088}]
+6   foo        autogen           2                 4            2017-06-19T00:00:00Z  2017-06-26T00:00:00Z                        [{5 stable.example:8088} {6 new.example:8088}]
+```
+
+Once you see that the old data node ID doesn't own any more shards, the anti-entropy job is complete.
+
+### Scenario 2: Replacing a machine that is running a data node
```
e-dard: I would also advocate using `update-data` in this case. It should be possible to run `update-data` on a machine that's still functioning in the cluster. The advantage of course is that the new machine gets the same node ID as the one that it's replacing and so the exact set of shards are guaranteed to be pulled onto the new machine.

content/enterprise_influxdb/v1.3/guides/anti-entropy.md
```diff
@@ -0,0 +1,138 @@
...
+$ influxd-ctl remove-data retired-data.example:8088
+[I] 2017-06-22T21:33:19Z Error: dial tcp retired-data.example:8088: getsockopt: connection refused. Retrying after 20ms
+[I] 2017-06-22T21:33:19Z Error: dial tcp retired-data.example:8088: getsockopt: connection refused. Retrying after 400ms
+Removed data node at retired-data.example:8088
+```
+
+It's common to see some intermittent network errors when removing a data node.
+However, if the command only prints errors and doesn't say that it removed the data node, use `influxd-ctl remove-data -force ADDRESS` to ensure that the node is removed from the cluster.
+
+Finally, the anti-entropy service will begin to copy shards to the new owner within the configured `anti-entropy.check-interval` duration.
```
e-dard: Might be worth mentioning that the copies will begin within that interval, but for large clusters it may take some time since the number of shards copied at any one time is limited by `anti-entropy.max-fetch`.

mark-rushakoff:
Thanks for the feedback and clarifications @e-dard.

content/enterprise_influxdb/v1.3/guides/anti-entropy.md
```diff
@@ -0,0 +1,112 @@
...
+1   _internal  monitor           2                 1            2017-06-22T00:00:00Z  2017-06-23T00:00:00Z  2017-06-30T00:00:00Z  [{5 stable.example:8088}]
+2   _internal  monitor           2                 1            2017-06-22T00:00:00Z  2017-06-23T00:00:00Z  2017-06-30T00:00:00Z  [{4 offline.example:8088}]
+4   _internal  monitor           2                 3            2017-06-23T00:00:00Z  2017-06-24T00:00:00Z  2017-07-01T00:00:00Z  [{5 stable.example:8088}]
+5   _internal  monitor           2                 3            2017-06-23T00:00:00Z  2017-06-24T00:00:00Z  2017-07-01T00:00:00Z  [{4 offline.example:8088}]
+6   foo        autogen           2                 4            2017-06-19T00:00:00Z  2017-06-26T00:00:00Z                        [{5 stable.example:8088} {4 offline.example:8088}]
+```
+
+We will replace the offline node with a new data node.
+Follow the [Data Node Installation Guide](/enterprise_influxdb/v1.3/production_installation/data_node_installation/) to start the new data node's `influxdb` service, but do not join the data node to the cluster.
```
e-dard: I would embolden or italicise `...but do not join the data node to the cluster.`. It's vital this is adhered to, or the `update-data` command will not work correctly.

e-dard:
@mark-rushakoff LGTM 👍

PR #1214, opened 2017-07-24
Remove relative url path

Ultimation:
This matches the rest of the document and means that clicking the link on any site other than influx's will work


timhallinflux:
Thanks @Ultimation

PR #1233, opened 2017-08-04
Update config.md

timhallinflux:
CQ and HTTP changes.


timhallinflux:
CQ
query_stats_enabled was missing from final config description...for 1.3

HTTP 
was missing the max_body_size

Issue #1259, opened 2017-09-06
Example results in 'invalid TICKscript:...'


DaveBender:
###### URL for relevant page?
https://docs.influxdata.com/kapacitor/v1.3/nodes/http_out_node/

###### What products and version are you using?
Kapacitor 1.3.1 (git: master 3b5512f7276483326577907803167e4bb213c613)

###### Where did you look before opening the issue?
At the Kapacitor documentation.

I'm trying to get a simple TICKscript working using the HTTPOutNode.  I copied the example provided on the documentation page into a file on my machine, then ran the kapacitor define command to create the Task, but got 

> invalid TICKscript: line 2 char 10: no method or property "window" on *pipeline.StreamNode

The script in the example is:

```

stream
        |window()
            .period(10s)
            .every(5s)
        |top('value', 10)
        //Publish the top 10 results over the last 10s updated every 5s.
        |httpOut('top10')

```
Seems like the example should work.

timhallinflux:
This will be published overnight -- 9/11/2017.

PR #1262, opened 2017-09-11
fixing example, added info

karel-rehor:
fixing issue 1259.  Added missing node in example. Added tip on trailing slashes and link to examples in repository.


karel-rehor:
Fixes issue #1259

PR #1265, opened 2017-09-12
changed holt_winters examples and descr. Fixes #1242

### LINKED ISSUE #1242: Add description on what holtwinters() node expects as arguments
###### URL for relevant page?
https://docs.influxdata.com/kapacitor/v1.3/nodes/where_node#holtwinters

kapacitor v1.3.1

https://docs.influxdata.com/influxdb/v1.3/query_language/functions/#holt-winters
https://github.com/influxdata/kapacitor/blob/v1.3.1/vendor/github.com/influxdata/influxdb/influxql/functions.go#L458
https://github.com/influxdata/kapacitor/blob/v1.3.1/vendor/github.com/influxdata/influxdb/influxql/functions.go#L416

The description in the kapacitor manual (4 arguments) does not fit to the linked influxql manual (3 arguments).

The new description could look like this (i did not get what interval does...):

### HoltWinters

Compute the holt-winters (https://docs.influxdata.com/influxdb/latest/query_language/functions/#holt-winters) forecast of a data set.  With `field` as as data source, `h` as how many datapoints should be generated, `m` as how many dapaoints long the season is and `interval` FIXME

```javascript
node|holtWinters(field string, h int64, m int64, interval time.Duration)
```

Returns: [InfluxQLNode](/kapacitor/v1.3/nodes/influx_q_l_node/)

karel-rehor:
Interesting lesson on forecasting.  Fixes #1242

### DIFF
--- a/content/kapacitor/v1.3/nodes/where_node.md
+++ b/content/kapacitor/v1.3/nodes/where_node.md
@@ -10,9 +10,9 @@ menu:
     parent: nodes
 ---
 
-The [WhereNode](/kapacitor/v1.3/nodes/where_node/) filters the data stream by a given expression. 
+The [WhereNode](/kapacitor/v1.3/nodes/where_node/) filters the data stream by a given expression.
 
-Example: 
+Example:
 
 
 ```javascript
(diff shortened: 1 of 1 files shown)


karel-rehor:
In the diff I noticed that my editor changed some CRLF characters at the ends of lines.  This probably should not be an issue.

Issue #1295, opened 2017-10-16
Escaping documentation for line protocol fields is not correct


danielnelson:
The line protocol documentation reads:

> For string field values use a backslash character `\` to escape:
> - double quotes `"`

However, according to the current implementation in InfluxDB, backslashes also need to be escaped.  This makes another line in the section incorrect:
> Line Protocol does not require users to escape the backslash character `\`.

###### URL for relevant page?

https://docs.influxdata.com/influxdb/v1.3/write_protocols/line_protocol_reference/#special-characters

###### What products and version are you using?

InfluxDB 1.3.4

###### Where did you look before opening the issue?

https://github.com/influxdata/influxdb/issues/8728

russorat:
I went through and tested all the inputs on that page and they all work in InfluxDB 1.3.6. I have added an additional example of how backslashes are interpreted by InfluxDB.

danielnelson:
If backslash is the last character then you must escape it, to avoid escaping the quote:

`temperature_str="too hot cold\"`

sqwishy:
The documentation states that you need to escape the space character on tag values.

> For tag keys, tag values, and field keys always use a backslash character \ to escape:
>    commas ,
>    equal signs =
>    spaces

https://github.com/influxdata/docs.influxdata.com/pull/1309 states that influx won't complain if you escape backslash characters, but this doesn't seem to be the case with tag values at all. If you try to escape them, it ignores that.

```
> insert foo,key=hello\ world v=0
> select * from foo;
name: foo
time                key         v
----                ---         -
1517956635336335149 hello world 0

> insert foo,key=\hello\ world v=0
> insert foo,key=\\hello\ world v=0
> select * from foo;
name: foo
time                key           v
----                ---           -
1517956635336335149 hello world   0
1517956648057229075 \hello world  0
1517956650713189940 \\hello world 0
```

This seems to throw things out of whack when it thinks your \ is escaping a space or something that isn't supposed to be escaped. Because you can't escape the escape character to get it to treat it like a regular character...

```
> insert foo,key=\\hello\ world\ v=0
ERR: {"error":"unable to parse 'foo,key=\\\\hello\\ world\\ v=0': invalid tag format"}
> insert foo,key=\\hello\ world\\ v=0
ERR: {"error":"unable to parse 'foo,key=\\\\hello\\ world\\\\ v=0': invalid tag format"}
```

Am I missing something or do backslash characters just not work at all for tag values.

[...]

PR #1309, opened 2017-11-02
adding additional examples for backslashes

russorat:
Attempting to address https://github.com/influxdata/docs.influxdata.com/issues/1295

image

### DIFF
--- a/content/influxdb/v1.3/write_protocols/line_protocol_tutorial.md
+++ b/content/influxdb/v1.3/write_protocols/line_protocol_tutorial.md
@@ -370,7 +370,33 @@ For string field values use a backslash character `\` to escape:
 weather,location=us-midwest temperature="too\"hot\"" 1465839830100400200
     ```
 
-Line Protocol does not require users to escape the backslash character `\`.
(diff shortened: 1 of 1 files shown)


sqwishy:
@russorat The documentation contradicts itself. If I want to insert `foo\\\bar`. Since the documentation states that I don't have to escape my backslash, that string should insert three backslashes. But inserts two instead. If I want to insert three backslashes, I have to escape at least two of them.

There are other cases where you need to escape a backslash if it is near another escape character. See @danielnelson's comment on https://github.com/influxdata/docs.influxdata.com/issues/1295.

russorat:
@sqwishy i agree the documentation is a little confusing here (and so is the behavior). If you want a single backslash, you can use \ or \\. If you want two backslashes, you can use \\\ or \\\\. and so on.

So the statement "I don't have to escape my backslash" is incorrect. The statement should be "For a single backslash, Influx treats single and double backslashes the same. If you need to insert multiple backslashes, they should be escaped." We will also need a few more details to address #1295 .

PR #1312, opened 2017-11-08
Updates to upgrade instructions for 1.3

### LINKED ISSUE #1304: Upgrading from Previous Versions does not mention expected downtime
###### URL for relevant page?
https://docs.influxdata.com/enterprise_influxdb/v1.3/administration/upgrading/

InfluxDB data 1.2.4_c1.2.5 and then InfluxDB data 1.3.1_c1.3.1

https://docs.influxdata.com/enterprise_influxdb/v1.3/administration/upgrading/
https://docs.influxdata.com/enterprise_influxdb/v1.3/about-the-project/release-notes-changelog/
https://github.com/influxdata/influxdb/blob/master/CHANGELOG.md

In the 'Upgrading from version 1.2.5 to 1.3.x', there is no mention of expected cluster disruption or downtime when, in reality, there must likely will be due to data node communication errors (due to a protocol change).
There is no mention of data-node-communication changes in the documents linked above.
For many people, upgrading single nodes in a cluster is assumed to be non-disruptive unless explicitly told otherwise (in their minds, the point of a cluster is that nodes can be dropped and restored without service downtime). Sadly, I am one of those people.
There is no mention that queries run during the cluster upgrade might provide incorrect data (as data nodes cannot communicate with each other to get all the data). There is no mention that Continuous Queries will continue to show incorrect data even after the upgrade for the period of time that the data nodes were being upgraded.
For extra points, you could provide an example of a Continuous Query and how to 'backfill' it between a couple of timestamps.

It would be amazing to know in advance that certain things are expected to fail when checking log files and/or executing queries when upgrading.

timhallinflux:
Updates to the install instructions to call out ordering of meta and data....also updating the instructions to call out removal/addition of the data nodes from the load balancer during the upgrade.

fixes #1304


content/enterprise_influxdb/v1.3/administration/upgrading.md
```diff
@@ -51,34 +61,82 @@ wget https://dl.influxdata.com/enterprise/releases/influxdb-meta_1.3.7-c1.3.7_am
...
+`influxd-ctl show` command.
+The [`influxd-ctl` tool](/enterprise_influxdb/v1.3/features/cluster-commands/) is available on all meta nodes.
+
+```
+~# influxd-ctl show
+
+Data Nodes
+==========
+ID	TCP Address		Version
+4	rk-upgrading-01:8088	1.3.7_c1.3.7   # 1.3.7_c1.3.7 = 👍
```
gunnaraasen: If meta nodes are upgraded first then I think the versions here need to be switched so the meta shows 1.3.7-c1.3.7 and the data nodes show 1.2.x-c1.2.x.

content/enterprise_influxdb/v1.3/administration/upgrading.md
```diff
@@ -51,34 +61,82 @@ wget https://dl.influxdata.com/enterprise/releases/influxdb-meta_1.3.7-c1.3.7_am
...
 
-#### Data node package download
+### Step 2: Install the 1.3.7 meta nodes packages
+
+#### Meta node package Install
+
 **Ubuntu & Debian (64-bit)**
 ```
-wget https://dl.influxdata.com/enterprise/releases/influxdb-data_1.3.7-c1.3.7_amd64.deb
+sudo dpkg -i influxdb-meta_1.3.7-c1.3.7_amd64.deb
```
gunnaraasen: Looks like one line got overwritten here. Should be the `wget` followed by the `dpkg`. Looks like the `yum` and both data node download and install instructions are also mangled.

```
wget https://dl.influxdata.com/enterprise/releases/influxdb-data_1.3.7-c1.3.7_amd64.deb
sudo dpkg -i influxdb-meta_1.3.7-c1.3.7_amd64.deb
```

content/enterprise_influxdb/v1.3/administration/upgrading.md
```diff
@@ -143,25 +191,34 @@ service influxdb restart
 sudo systemctl restart influxdb
 ```
 
-### Step 5: Confirm the upgrade
+### Step 6: Add the data node back into the load balancer
+
+Re-add the data node back into the load balancer to allow it to serve reads and writes.
+If this is the last data node to be upgraded, proceed to step 7.  
+
+Otherwise, return to step 1 for the data nodes and
```
gunnaraasen: We may also want to mention that data nodes still on 1.2.x will sometimes complain about hinted handoff until all nodes are updated.

content/enterprise_influxdb/v1.3/administration/upgrading.md
```diff
@@ -143,25 +191,34 @@ service influxdb restart
...
-4	rk-upgrading-01:8088	1.3.7_c1.3.7   # 1.3.7_c1.3.7 = 👍
+4	rk-upgrading-01:8088	1.3.7_c1.3.7   
 5	rk-upgrading-02:8088	1.3.7_c1.3.7
 6	rk-upgrading-03:8088	1.3.7_c1.3.7
 
 Meta Nodes
 ==========
 TCP Address		Version
-rk-upgrading-01:8091	1.3.7_c1.3.7
+rk-upgrading-01:8091	1.3.7_c1.3.7       # 1.3.7_c1.3.7 = 👍
```
gunnaraasen: The comment should probably be moved to the data nodes section.

content/enterprise_influxdb/v1.3/administration/upgrading.md
timhallinflux: I think you are comparing this to the install instructions?   We are working on the upgrade instructions.... The download and install and service start commands and instructions are separated in the upgrade doc.

content/enterprise_influxdb/v1.3/administration/upgrading.md
timhallinflux: added!

timhallinflux:
fixes 1307

content/enterprise_influxdb/v1.3/administration/upgrading.md
gunnaraasen: Even for upgrades, the package file with the newer version needs to be downloaded before it can be installed. If a package repo was used for install/upgrade instead of the package file, then the newer version package would not need to be explicitly downloaded because it would be automatically downloaded by the package manager. However, these instructions are for an installing from a package file so the file needs to be downloaded.

PR #1349, opened 2017-12-15
Telegraf 1.5

stevebang:
Load new Telegraf 1.5 docs, including merging of input and service plugin folders and new changelogs, etc.


stevebang:
Need a few more changes.

Issue #1356, opened 2017-12-19
Is GOMAXPROCS an environment variable or a configuration file variable or both?


omki2005:
This link outlines the list of environment variables that influxdb supports :- 

https://docs.influxdata.com/influxdb/v1.4/administration/config/#environment-variables

This link outlines the list of global configuration options in the config file that influxDB supports :- 

https://docs.influxdata.com/influxdb/v1.4/administration/config/#global-options

Two of the three config file global variables are output by influxd when creating a default config file namely "reporting-disabled" and "bind-address" so clearly the list above is referring to config file options.

However, this comment by a contributor to influxDB suggests that GOMAXPROCS is an environment variable option only and not a config file option :- 

https://github.com/influxdata/influxdb/issues/9249

The documentation seems to be incorrect in that regard.

jwilder:
From #1357:

The GOMAXPROCS environment variable is listed as a global config options implying that it can be set via the config file. This is not accurate, you can only set it via the GOMAXPROCS environment variable.

It should be updated to reflect that you can't set it via the config and possibly moved to https://docs.influxdata.com/influxdb/v1.4/administration/config/#environment-variables to avoid confusion.

stevebang:
GOMAXPROCS was moved to new environment variables section. Corrections made.

PR #1358, opened 2017-12-19
GOMAXPROCS only configurable via environment variable

omki2005:
Fixes https://github.com/influxdata/docs.influxdata.com/issues/1356


timhallinflux:
@omki2005 we are going to address this differently.  See #1361

PR #1376, opened 2018-01-10
Update different options for per-database backups

dprittie:
When reading the current documentation my colleagues and I got pretty confused and initially were under the impression that no form of incremental backup was supported.

I think this amendment makes the document a bit clearer and more accurate.

One thing that may be worth considering is that I have only updated the v1.4 documentation and perhaps it would be appropriate to update more than just that version, which I am happy to do. As far as I can tell the amendment I have made is accurate for all versions of influxdb since v0.10.0. The first commit I found in influxdata/influxdb with incremental backups was df1aeee7

### DIFF
--- a/content/influxdb/v1.4/administration/backup_and_restore.md
+++ b/content/influxdb/v1.4/administration/backup_and_restore.md
@@ -9,12 +9,10 @@ menu:
 
 ## Backups
 
-InfluxDB has the ability to snapshot an instance at a point-in-time and restore it.
-All backups are full backups.
-InfluxDB does not yet support incremental backups.
 There are two types of data to backup, the metastore and the metrics themselves.
 The [metastore](/influxdb/v1.4/concepts/glossary/#metastore) is backed up in its entirety.
 The metrics are backed up per-database in a separate operation from the metastore backup.
+Per-database backups can be full, incremental (since a RFC3339 formatted time), or for a specific shard ID.
 
 > **Note:** Backups are not interchangeable between OSS InfluxDB and [InfluxEnterprise](/enterprise/latest/).
 You cannot restore an OSS backup to an InfluxEnterprise data node, nor can you restore


aanthony1243:
@dprittie  Thanks for offering some feedback + clarification!  The term 'incremental'  backup in Devops nomenclature is that you can take an existing backup datafile and extend it with anything new occurring since the previous backup operation.  

The 'since'  parameter makes it feel like an incremental backup in a way, but each backup in this context will be self-contained and cannot be 'stacked', e.g.,  if you take a backup 'since'  12/1/2017  and then later take a backup 'since' 1/1/2018,  then you may either restore to the state of the data taken on 12/1/2017  or 1/1/2018, the two data files cannot be combined.   

the semantic of 'since' is "Take a full backup, but discard all data prior to this date".   

Perhaps we can amend your suggestion to :   "Per-database backups can be for the entire contents of the database, or limited only to data files that have been changed since a RFC3339 formatted time. Alternatively, a backup may be limited to a specific retention policy or shard ID."

dprittie:
@aanthony1243 Thank you for the feedback. Given that backups cannot be stacked your amendment makes sense to me, although I would be tempted to change "Alternatively, a backup" to "Additionally, a backup" as I think the options in question can be used in conjunction with --since. Would you like to make that change, or should I?

In light of my lack of understanding on this topic, I would be grateful if you could help me with a follow up question regarding incremental/stacking. Let's say that every week I backup a database using the --since parameter from the beginning of the week and store it in /var/backup/influx/db/week/# and immediately afterwards I backup the meta store to /var/backup/influx/meta
The database I am backing up is the only db I care about, and whilst I keep 10 weeks of db backups, I only ever have one meta store backup (the latest).
Then disaster strikes and I lose all my data on this influx server.
So I set up a brand new server and begin restoring data from the oldest week to the newest week, but each time using the latest meta store backup.
Are you saying that the only data I will end up with will be from /var/backup/influx/db/week/10 and I will have lost everything from 1-9?

aanthony1243:
@dprittie yes, that's the current functionality.  You'd want to avoid the 'since' keyword entirely for your situation.  The interpretation of 'since' is "throw out anything earlier than this date".  Also the date is read from the last time a file was opened on disk, not by the timestamps in the data.   So secondary operations such as compaction may change the modified time on every file.  

We have some new backup features staged for the next release, which should be coming soon.  Your backup plan may be more do-able with the new features, so be sure to check back.  

@stevebang   can make the change to the 1.4 docs as he sees appropriate, he's the lead on our docs effort now.

PR #1408, opened 2018-02-01
Propose small spelling changes

jeis2497052:
also suggesting the removal of some trailing white space if that helps?


timhallinflux:
thanks @jeis2497052

PR #1443, opened 2018-03-07
Bonitoo sp6 - Kapacitor Open Source Security

karel-rehor:
Contains new document content/kapacitor/v1.4/administration/security.md

Stage is here.

http://ec2-54-153-106-230.us-west-1.compute.amazonaws.com:1414/kapacitor/v1.4/administration/security/

> Note: that at one point the document content/influxdb/v1.4/administration/security.md was committed by mistake.  After realizing the error I reverted it back to the version in master.  Just to be sure that this merge will not overwrite or lead to conflicts with any other changes.


content/kapacitor/v1.4/administration/security.md
```diff
@@ -0,0 +1,450 @@
...
+* [Overview](#overview)
+* [Secure InfluxDB and Kapacitor](#secure-influxdb-and-kapacitor)
+* [Kapacitor Security](#kapacitor-security)
+* [Secure Kapacitor and Chronograf](#secure-kapacitor-and-chronograf)
+
+# Overview
+
+This document covers the basics of securing the open-source distribution of
+Kapacitor.  For information about security with Enterprise Kapacitor see the
+[Enterprise Kapacitor](/enterprise_kapacitor/v1.3/) documentation.
```
nathanielc: Update link for 1.4

content/kapacitor/v1.4/administration/security.md
```diff
@@ -0,0 +1,450 @@
...
+environment variables or over Kapacitor's HTTP API.  
+
+Storing passwords and other sensitive information in a configuration file on
+disk is discouraged in production environments.  However Kapacitor cannot
+successfully start without communications to a default InfluxDB server.  It is
+therefore advised to use the configuration file to boot Kapacitor, and then to
+make use of the HTTP API when resetting sensitive values, after they have been
+redacted from the configuration file on the storage volume.  Note that if
+Kapacitor needs to be rebooted, the parameters in the file will need to be
+temporarily restored.
```
nathanielc: This is not true, there is no security advantage using the HTTP API over the configuration file. Both store the credentials in plain text on disk.

content/kapacitor/v1.4/administration/security.md
```diff
@@ -0,0 +1,450 @@
...
+therefore advised to use the configuration file to boot Kapacitor, and then to
+make use of the HTTP API when resetting sensitive values, after they have been
+redacted from the configuration file on the storage volume.  Note that if
+Kapacitor needs to be rebooted, the parameters in the file will need to be
+temporarily restored.  
+
+An overview of Kapacitor configuration is provided in the
+[Configuration](/kapacitor/v1.4/administration/configuration/) document.
+
+### Note on HTTP API Configuration and Restarting Kapacitor
```
nathanielc: I think this section should go toward the end. At this point in the doc I still haven't been instructed on how to actually enable TLS on Kapacitor. I'd rather these notes be placed towards to end and the content the user is trying to learn be towards the beginning.

content/kapacitor/v1.4/administration/security.md
```diff
@@ -0,0 +1,450 @@
...
+  urls = ["https://localhost:8086"]
+  timeout = 0
+  # Absolute path to pem encoded CA file.
+  # A CA can be provided without a key/cert pair
+  #   ssl-ca = "/etc/ssl/influxdata-selfsigned-incl-pub-key.pem"
+  # Absolutes paths to pem encoded key and cert files.
+     ssl-cert = "/etc/ssl/influxdb-selfsigned.crt"
+     ssl-key = "/etc/ssl/influxdb-selfsigned.key"
+...
+   insecure-skip-verify = true
```
nathanielc: My preference is that the examples here are production ready examples and that we simply note that this should be switched off for local testing.

That way some that is blindly copy pasting gets a secure configuration.

content/kapacitor/v1.4/administration/security.md
karel-rehor: Updated.  However, on local doc build and stage leads to a broken link. Enterprise Kapacitor will need to be updated, accordingly

content/kapacitor/v1.4/administration/security.md
karel-rehor: Removed this paragraph. 

As an aside, on previous "Enterprise" projects, on which I've worked (e.g. at HP). _password_ properties in configuration files were discouraged or were hashed.  The service on boot when reading the configuration file would detect plain text passwords and overwrite them with a hash value.

content/kapacitor/v1.4/administration/security.md
karel-rehor: Move the note on Restarting Kapacitor to end of Kapacitor TLS configuration section and just before discussion of Chronograf integration.

content/kapacitor/v1.4/administration/security.md
karel-rehor: Yes, it is better to have production ready examples, so that end user do not inadvertently introduce a security hole by simply following the examples, without reading the text.  Updated to `insecure-skip-verify = false` also update this value in json example further down.

karel-rehor:
I've just added `content/enterprise_kapacitor/v1.4` to this PR - including an update to `data/default_versions.toml`.  

@nathanielc do you have the CHANGELOG.md for enterprise kapacitor for the 1.4 release?  Can you send it by email? The content of the document `content/enterprise_kapacitor/v1.4/release-notes-and-changelog.md` will need to be updated, however I have no access to the enterprise_kapacitor project on github.

nathanielc:
@karel-rehor The Enterprise Kapacitor 1.4 release was only a bug fix release + the OSS features. So the changelog is the same as the OSS Kapacitor 1.4 release.

PR #1470, opened 2018-03-26
Feature/view in chronograf, resolves #1315

### LINKED ISSUE #1315: All examples should have an "Open in Chronograf" button
Just an idea to make working with examples easier. If the user has a Kapacitor up and running on their machine, it would be cool to have a button next to any queries or kapacitor scripts to "Open in Chronograf" which would open a new page at localhost:8888 with the query or script populated.

see the Elastic docs for how something like this might work: https://www.elastic.co/guide/en/elasticsearch/reference/current/_cluster_health.html

sanderson:
To include a "View in Chronograf" button, immediately after a code block, add an html `div` with the `view-in-chronograf` class and pass the query in the `data-query` attribute.

_Be aware of single vs double quotes when passing the query. I recommend wrapping the entire query in single quotes to preserve the double quotes in the query and to avoid closing the string too early._

Examples:

````md
```js
SHOW DATABASES
```
<div class='view-in-chronograf' data-query='SHOW DATABASES'></div>
````

````md
```js
SELECT * FROM "mult" WHERE "A" * 10 >= 20
```
<div class='view-in-chronograf' data-query='SELECT * FROM "mult" WHERE "A" * 10 >= 20'></div>
````

This will append the "View in Chronograf" button to the code block along with a Chronograf URL settings button (the cog).

screen shot 2018-03-26 at 8 57 11 am

When the "View in Chronograf" button is clicked, it'll open the Chronograf data explorer in a new window with the query pre-populated. To change the Chronograf URL, users can click the settings cog which will open a modal where they can define their own custom Chronograf URL. This gets saved as a cookie and will persist across their session.

_There currently isn't an expiration set on the cookie, but it can be easily added if we decide we need one._

### CODE REVIEW

File: static/js/docs-chronograf-url.js [hunk 1]
```diff
@@ -0,0 +1,69 @@
...
+
+// Set/Update the cookie
+function updateChronografUrl ( chronograf_url ) {
+  Cookies.set(chronograf_cookie_name, chronograf_url)
+}
+
+// Get/set the Chronograf URL
+function getChronografUrl() {
+  var chronografUrl = ( Cookies.get('influx-chronograf-url') === undefined ? chronograf_default_url : Cookies.get('influx-chronograf-url') )
+  return chronografUrl
```
russorat: they may also have a different source number (other than 1). Not sure if this allows them to save that as well.

sanderson: The source number is included in the Chronograf URL settings modal. It's not a separate field, but can be changed in URL itself.

image

PR #1487, opened 2018-03-30
Feature - add robots "noindex, nofollow" meta tag on legacy docs

sanderson:
The logic here pulls the project name (`influxdb`, `influxdb_enterprise`, `chronograf`, etc.) out of the url and checks it agains the project's semver defined `data/default_versions.toml`. If the projects semver doesn't exist in the page's URL, the robots tag is set to `noindex, nofollow`. If it does exist in the URL or the current page is the home page, it sets it to `index, follow`.

This should help to prevent users being led to old docs from Google search results.


russorat:
when this is checked in, someone should log into our google search console and request a recrawl of the docs.

stevebang:
This will tell Google to index only the latest versions of our products. The Docsearch will still index all pages for all versions, so customers on earlier versions can use the docsite's search to find earlier docs if needed.

PR #1500, opened 2018-04-03
added docs home link in top nav with responsive styles, resolves #1493

### LINKED ISSUE #1493: Add a link to the docs homepage in the top nav
There currently isn't a link in the top nav that will take visitors back to the home page of the docs. The link there takes you back to influxdata.com.

sanderson:
This adds a "Docs" link in the top nav that links back to the docs homepage. On larger screens, it looks like this:

screen shot 2018-04-03 at 4 23 54 pm

On smaller screens, it scales down to this:

screen shot 2018-04-03 at 4 24 42 pm


sanderson:
One thing I have noticed is that the Klavika font, the font used in the logo and for "Docs" here, doesn't display in Safari. I think the issue is that the current Klavika font is being rendered using an `.otf` (OpenTypeFont) font format. Chrome handles `.otf` files well, but other browsers don't. We may need to purchase a web license for this font to get the proper font files that are supported by all modern browsers.

PR #1529, opened 2018-04-10
Add CA certificate and HTTPS client certificate verification support for InfluxDB

gentstr:
Documentation for influxdata/influxdb#9421

###### Required for all non-trivial PRs
- [x] Sign [CLA](https://influxdata.com/community/cla/)


jsternberg:
@gentstr @stevebang this feature was never merged so this doesn't work. We have an open PR for it so can you revert this and then maybe we can reapply the same documentation if we merge the PR? Thanks.

PR #1542, opened 2018-04-16
Fix http://localhost:1313 links

LukeShu:
A couple of links are rooted at "http://localhost:1313/" instead of "/",
which means they work with the local `hugo server`, but not when deployed
at https://docs.influxdata.com/


sanderson:
Nice catch @LukeShu. Thanks for the fix.

PR #1600, opened 2018-05-07
Add Kapacitor 1.5 autogen docs

nathanielc:
There are two commits in this PR. The first copies over the 1.4 nodes, the second applies the autogen changes.


sanderson:
We're moving away from the autogen'd docs so I'm not going to merge this, but I cherry-picked all the important changes. Thanks for submitting these.

PR #1603, opened 2018-05-08
fix typos in config.md

Bigjaked:
deleted extra "those" and "the"


sanderson:
Thanks @Bigjaked! Nice catch.

PR #1606, opened 2018-05-09
Added Kapacitor recoverable alert field option

sanderson:


sanderson:
@veereshhosakere, the explanation and details of the `Recoverable` field option are in this PR, but I don't have any user stories or in-the-wild use-cases for it. Is there any insight you could provide?

PR #1609, opened 2018-05-10
Refactored the Kapacitor template tasks doc

sanderson:


content/kapacitor/v1.5/working/template_tasks.md
```diff
@@ -53,25 +55,57 @@ stream
...
+var varName = dataType: defaultValue
+var varName = [*]
 ```
-kapacitor define-template generic_mean_alert -tick path/to/above/script.tick
+
+_View the [literal value types](/kapacitor/v1.5/tick/syntax/#types) section of the
+TICKscript syntax article for information about available data types._
+
+#### Optional variables
+In some cases, a templated task may be used for tasks that do not require a values for all templated variables.
```
sanderson: Remove "a" from "required a values"

PR #1619, opened 2018-05-17
Add missing argument to backup_and_restore.md

petetnt:
The `restore` example was missing the `[ -newdb <newdb_name> ]` optional argument from the usage example.


stevebang:
Thanks, @petetnt for catching this

PR #1624, opened 2018-05-22
Update licenses.md

Jarf10:
Fixed typo


sanderson:
Nice catch. Thanks for the PR.

PR #1655, opened 2018-06-04
Added information about differences between CQs and TICKscript results

sanderson:


content/kapacitor/v1.5/guides/continuous_queries.md
```diff
@@ -157,3 +157,39 @@ Kapacitor is a powerful tool, if you need more power use it.
 If not keep using CQs until you do.
 For more information and help writing TICKscripts from InfluxQL queries take a looks at these [docs](https://docs.influxdata.com/kapacitor/latest/nodes/influx_q_l_node/) on the InfluxQL node in Kapacitor.
```
stevebang: "take a looks at"? Also, a comma is needed after queries.

content/kapacitor/v1.5/guides/continuous_queries.md
```diff
@@ -157,3 +157,39 @@ Kapacitor is a powerful tool, if you need more power use it.
 If not keep using CQs until you do.
```
stevebang: Needs a comma after "If not".

sanderson: I'm guessing it won't let you select it because that line isn't part of this PR. But it does need to be updated.

PR #1660, opened 2018-06-05
Kapacitor/two measurement alerts

sanderson:


content/kapacitor/v1.5/guides/two-measurement-alert.md
```diff
@@ -0,0 +1,131 @@
+---
+title: Creating alerts based on two measurements
```
stevebang: Change title to better describe alerts being triggered. Something like, "Triggering alerts based on differences between measurements"?

PR #1661, opened 2018-06-06
Update GitHub Oauth instructions to use token-secret command line option

chadothompson:
I found issue in the GitHub OAuth provider documentation:  the test case

a) . I configured using the variable `--token_secret` per documentation.  Was getting errors.

b).  Updated to follow convention with other variables ( `--token-secret`), and things seemed to work.

Test case (OS X from `brew install chronograf`):

```
    <string>homebrew.mxcl.chronograf</string>
    <key>ProgramArguments</key>
    <array>
      <string>/usr/local/opt/chronograf/bin/chronograf</string>
      <string>--public-url=http://localhost:8888</string>
      <string>--github-client-id=REDACT</string>
      <string>--github-client-secret=REDACT</string>
      <string>--token-secret=REDACT</string>
    </array>
```


chadothompson:
Also, see 

https://github.com/influxdata/chronograf/blob/49e5c88fb00ff87f478b12efd8255a8991d64e84/server/server.go

(Long = `token-secret`)

Issue #1665, opened 2018-06-07
Update `max-concurrent-compactions` in 1.5 


timhallinflux:
###### URL for relevant page?
http://docs.influxdata.com/influxdb/v1.5/administration/config/
and
http://docs.influxdata.com/enterprise_influxdb/v1.5/administration/configuration/#max-concurrent-compactions-0

###### What products and version are you using?
1.5.x

The current text reads:
The maximum number of concurrent full and level compactions that can run at one time. A value of 0 results in runtime.GOMAXPROCS(0) used at runtime – which means use all processors. This setting does not apply to cache snapshotting.

This was true to 1.3.x

However, the algorithm has changed.  
A value of 0 results in (processors / 2) -- if the reported value is less than 8.  
If the reported value is 8 or more, the value of 4 is used.

If you wish to use MORE processors for compactions, this setting can be used to increase the amount of processors used.  This is highly recommended for clusters with a large number of processors and a large amount of data ingestion.

e-dard:
@timhallinflux, I noticed this too and mentioned to @stevebang on our channel. I’ll look over the resulting PR.

stevebang:
Closing -- new description of changes in #1669 .

PR #1669, opened 2018-06-08
Modify max-concurrent-compactions description

stevebang:
This clarifies the behavior of this setting. Based on discussions with @e-dard.


stevebang:
Resolves #1665.

timhallinflux:
I think we need to highlight this change in the release notes -- and further, we need to embellish this point somewhere in troubleshooting?   Large machines (greater than 8 cores) will restrict you to 4 cores for compaction -- this may be insufficient to achieve maximum compaction based on your workload. Increasing this setting along with ```cache-snapshot-memory-size``` to something larger as well (like 512MB ... 52428800).

stevebang:
Merging and will address which earlier versions/release notes need to be modified.

PR #1680, opened 2018-06-20
added percentage calculation guide, extended view in chronograf funct…

### LINKED ISSUE #1572: Cannot perform a simple query to find a percentage
Hello,

I am attempting to perform a query to find a simple percentage, but have been struggling.

I'm looking for something as simple as this:
`SELECT gain_amount / sum(gain_amount) FROM sells`
But get the error that I cannot mix aggregate and non-aggregate queries. This is confusing to me because the following works:

`SELECT gain_amount / 123 FROM sells`
which seems like the same thing to me, as sum only returns one value, 123.

I've also tried:
`SELECT gain_amount / (SELECT sum(gain_amount) FROM sells ORDER BY DESC LIMIT 1) FROM sells`
which fails as in issue #5684 below

I saw the sub query implementation as I showed below and have tried to get extremely creative with those to no avail. I've tried the following any **many** different spins on it:

`SELECT gain_amount / "sum" FROM (SELECT sum(gain_amount) FROM sells)`
Which returns an empty set, i'm assuming because it doesn't know where my first gain_amount is coming from due to the fact that my sub-query has to be after the FROM keyword.

Any help here? I must be missing something simple.

v 1.5.1

I found https://github.com/influxdata/influxdb/issues/5684, which was closed with https://github.com/influxdata/influxdb/issues/4619, but does not seem to resolve the issue at all.

sanderson:
…ionality, resolves #1572


sanderson:
Will add this to the 1.6 branch when approved.

content/influxdb/v1.5/guides/calculating_percentages.md
```diff
@@ -0,0 +1,115 @@
+---
+title: Calculating Percentages in a Query
```
stevebang: Change title to sentence-style and plural of  "Calculating percentages in queries"

PR #1684, opened 2018-06-22
Update/document sensu metadata

sdomino:
This is to accompany https://github.com/influxdata/kapacitor/pull/1981

### DIFF
--- a/content/kapacitor/v1.5/event_handlers/sensu.md
+++ b/content/kapacitor/v1.5/event_handlers/sensu.md
@@ -43,10 +43,11 @@ The following Sensu event handler options can be set in a
 [handler file](/kapacitor/v1.5/event_handlers/#handler-file) or when using
 `.sensu()` in a TICKscript.
 
-| Name     | Type            | Description                                                                |
-| ----     | ----            | -----------                                                                |
-| source   | string          | Sensu source for which to post messages.                                   |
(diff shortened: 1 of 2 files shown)


content/kapacitor/v1.5/event_handlers/sensu.md
```diff
@@ -67,6 +71,7 @@ options:
   .sensu()
     .source('Kapacitor')
     .handlers('handler1-name', 'handler2-name')
+    .metadata({'string':'value', 'int':1})
```
mdelaney: According to the documentation, tick script syntax doesn't support this. https://docs.influxdata.com/kapacitor/v1.5/tick/syntax/#types shows the supported data types. Indeed, when I test this I get a syntax error.

My tick script fragment is
```
        .sensu()
        .source('my-source')
        .handlers('myhandler')
        .metadata({'playbook':'http://foo.bar', 'occurrences':3})
```
Which results in an error like the following when you define it
```
invalid TICKscript: parser: unexpected unknown state, last char: '{' line 38 char 19 in ".metadata({'playbook". expected: "number","string","duration","identifier","TRUE","FALSE","==","(","-","!"
```

sanderson:
The original PR was closed in favor of another approach to adding this meta data.

PR #1694, opened 2018-06-27
Add note about epoch rounding and clean up examples

stevebang:
Adds note about epoch time values returned rounded to the nearest whole unit. Also, did some layout cleanup of examples.

Resolves ##919.


content/influxdb/v1.5/guides/querying_data.md
```diff
@@ -125,31 +128,45 @@ returns:
...
+
+##### Example: Get epoch in seconds:
+
 ```bash
 curl -G 'http://localhost:8086/query' --data-urlencode "db=mydb" --data-urlencode "epoch=s" --data-urlencode "q=SELECT \"value\" FROM \"cpu_load_short\" WHERE \"region\"='us-west'"
 ```
 
+##### Example: Rounding of epoch time value
+
+If a user had a timestamp of  `1970-01-01T00:00:00.000051Z` and queried back for it with `epoch=ms`, the result is time value of `0`.
```
sanderson: I wonder if a second example of a timestamp rounded up here would help drive the point home.

Issue #1716, opened 2018-07-11
Add influx_inspect deletetsm command to 1.6


sanderson:
###### URL for relevant page?
https://docs.influxdata.com/influxdb/v1.6/tools/influx_inspect/

The `deletetsm` command isn't documented in the influxdb 1.6 docs. In the 1.5 docs, it's documented, but needs a anchored link at the top of the page.

sanderson:
Note that when specifying the path for the deletetsm command, wildcard paths are supported.

Single shard:
```
./influx_inspect deletetsm -sanitize /influxdb/data/location/autogen/1384/*.tsm
```

All shards int he db:  
```
./influx_inspect deletetsm -sanitize /influxdb/data/location/autogen/*/*.tsm
```

PR #1724, opened 2018-07-18
Influxdb 1.6 deletetsm fix

stevebang:
Adds missing deletetsm to 1.6, adds missing anchor link, and note about wildcards.


stevebang:
Fixes #1716.

PR #1740, opened 2018-07-24
Chronograf predefined influxdb and kapacitor sources

sanderson:


content/chronograf/v1.6/administration/config-options.md
```diff
@@ -84,6 +84,14 @@ Default value: `/usr/share/chronograf/canned`
 
 Environment variable: `$CANNED_PATH`
 
+#### `--resources-path=`
+
+Path to directory of pre-canned dashboards, sources, kapacitors, and organizations
```
stevebang: Either say "canned" (the environment variable is CANNED_PATH") or, "predefined (or canned) ..." — ultimately, the goal is to get rid of "canned", but with it in the code, it needs too much work for now.

stevebang: "kapacitors"?  Do you mean Kapacitor connections?

sanderson: Do you mean replacing "pre-canned" with "canned." `--resources-path` is the name of the actual argument.

The description was pulled straight from the CLI help output, but it would probably be more clear if it were "Kapacitor connections." I'll change it.

PR #1747, opened 2018-07-27
Fix some typos and update example output

incous:
I found some typo and also update output example for version 1.6.x; just minor edit, easy for review


sanderson:
Thanks @incous!

PR #1755, opened 2018-08-03
Deleted repeated words, added and removed commas, renamed ampersand operator

paulinalei:
Hi, 

This is my first time doing open source contributions, so if I have made any mistakes, please let me know!

I wasn't too sure about how to do semantic linefeeds, so if anyone could explain or help me out with that, that would be great!

For the fixes I did, I will explain them here:
All of the fixes I made are in the syntax page: https://docs.influxdata.com/kapacitor/v1.5/tick/syntax/

Around example 5, I noticed that there was a sentence "Strings begin with either one or three single single quotation marks" and believe that "single" was repeated twice.

Around example 15, I noticed there was a sentence "In the where() method call the lambda expression" and believe it could be clearer if there was a comma after "call."

Around example 17, I noticed that there was a sentence "To access a property a period, ., is used before the identifier" and believe it would be clearer to have a comma after "property" and remove the ones after "period."

Around the section for expressions, I noticed there was a sentence "The ampersand operator “@” introduces a user defined function" but I believe this is the at operator and ampersand is "&." If I am incorrect in this naming convention, please let me know. I was unsure about this change.

### DIFF
--- a/content/kapacitor/v1.5/tick/syntax.md
+++ b/content/kapacitor/v1.5/tick/syntax.md
@@ -230,7 +230,7 @@ In Example 5 above the first two lines show the declaration of Duration types.  
 
 
 ##### Strings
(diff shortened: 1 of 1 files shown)


sanderson:
Hi @pauldix,

Thanks for the contribution! Semantic line breaks essentially mean 1 sentence per line, but if the sentence is excessively log, you can break a single sentence onto multiple lines. When the markdown is parsed and rendered as HTML, the newline will not be included. Paragraphs are separated by double newlines.

No all the markdown files are like this yet, but we're slowly working through them all.

paulinalei:
Hi @sanderson,

Thank you for explaining it!

PR #1761, opened 2018-08-06
Deleted repeat words, moved commands, renamed operator

paulinalei:
In the section "Node creation," I moved the command to after "(stream and batch)" for clarification.

In the section "Pipeline," "with" was repeated, so I removed one and the @ operator was named ampersand, so I changed it to "at."


sanderson:
Thanks @paulinalei!

Issue #1769, opened 2018-08-10
TICKScript in "Calculating rates across joined series" guide has errors


bunyk:
```
 |eval((lambda: "errors.sum" / ("views.sum" + "errors.sum")) * 100)
```

Gives error during definining task: invalid TICKscript: parser: unexpected lambda line 29 char 12

Probably it should be written as 
```
    |eval(lambda: "errors.sum" / ("views.sum" + "errors.sum") * 100)
```
But it gives errors during runtime:

ts=2018-08-10T13:46:47.662Z lvl=error msg="error evaluating expression" service=kapacitor task_master=main task=error_percent node=eval9 err="mismatched type to binary operator. got float * int. see bool(), int(), float(), string(), duration()"

So maybe it should be:
```
    |eval(lambda: "errors.sum" / ("views.sum" + "errors.sum") * 100.0)
```

At least that works for me. 

###### URL for relevant page?
https://docs.influxdata.com/kapacitor/v1.5/guides/join_backfill/

###### What products and version are you using?
Kapacitor
Version:                      1.5.0                         

###### Where did you look before opening the issue?
into the `/var/log/kapacitor/kapacitor.log`

PR #1770, opened 2018-08-10
Fix lambda in tickscript 

bunyk:
Fixes issue #1769

### DIFF
--- a/content/kapacitor/v1.5/guides/join_backfill.md
+++ b/content/kapacitor/v1.5/guides/join_backfill.md
@@ -227,7 +227,7 @@ errors
     |join(views)
         .as('errors', 'views')
     // Calculate percentage
-    |eval((lambda: "errors.sum" / ("views.sum" + "errors.sum")) * 100)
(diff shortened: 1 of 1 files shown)


sanderson:
Thanks for the fix @bunyk. This one was my fault. and I see what I did wrong. I was using parenthesis to force the order of mathematical operations to ensure that the result of the division would be multiplied by 100 rather than the denominator being multiplied by 100 before the division happened. The opening parenthesis was outside of the lambda expression when it should've been inside, like this:

```js
|eval(lambda: ("errors.sum" / ("views.sum" + "errors.sum")) * 100.0)
```

Your solution works as well since the order of mathematical operations is handled properly at runtime, so the division takes place first, then the result is multiplied by 100.

bunyk:
Your welcome.

Could you or someone else also help me with this https://community.influxdata.com/t/how-to-debug-join-node-in-kapacitor/5989 ? It is almost the same as the script in this tutorial, except that data are stored in same measurement tagged, and I want to compute percentage by tags, so I want to divide each datapoint by total sum across tag values, but in the end there is no data. :(

PR #1780, opened 2018-08-15
Update sample data table in data_exploration.md

SiavasFiroozbakht:
Properly formatted the subsample table.

### DIFF
--- a/content/influxdb/v1.6/query_language/data_exploration.md
+++ b/content/influxdb/v1.6/query_language/data_exploration.md
@@ -68,56 +68,16 @@ InfluxDB shell 1.4.x
 Next,  get acquainted with this subsample of the data in the `h2o_feet` measurement:
 
 name: <span class="tooltip" data-tooltip-text="Measurement">h2o_feet</span><br />
-\------------------------------------<br />
-time
-&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;
-<span class="tooltip" data-tooltip-text="Field Key">level description</span>
-&emsp;&emsp;&emsp;&emsp;
-<span class="tooltip" data-tooltip-text="Tag Key">location</span>
-&emsp;&emsp;&emsp;&emsp;
-<span class="tooltip" data-tooltip-text="Field Key">water_level</span><br />
-2015-08-18T00:00:00Z
-&emsp;&emsp;
-between 6 and 9 feet
-&emsp;&emsp;&thinsp;
-coyote_creek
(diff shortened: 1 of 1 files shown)


sanderson:
Thanks for this @SiavasFiroozbakht. The changes you propose definitely help, but I think it would be better to format this as an actual table rather than forcing spacing and line breaks. As we've been reviewing documentation, we've been updating this type of formatting which is an artifact of old documentation builds.

The main issue I see with what you propose is that on smaller screens, line wrapping still causing the information to be hard to read. I propose the following markdown table:

```
| time                                                                            | <span class ="tooltip" data-tooltip-text ="Field Key">level description</span>      | <span class ="tooltip" data-tooltip-text ="Tag Key">location</span>       | <span class ="tooltip" data-tooltip-text ="Field Key">water_level</span> |
| ------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------- | ------------------------------------------------------------------------- | -----------------------------------------------------------------------: |
| 2015-08-18T00:00:00Z                                                            | between 6 and 9 feet                                                                | coyote_creek                                                              | 8.12                                                                     |
| 2015-08-18T00:00:00Z                                                            | below 3 feet                                                                        | santa_monica                                                              | 2.064                                                                    |
| <span class="tooltip" data-tooltip-text="Timestamp">2015-08-18T00:06:00Z</span> | <span class ="tooltip" data-tooltip-text ="Field Value">between 6 and 9 feet</span> | <span class ="tooltip" data-tooltip-text ="Tag Value">coyote_creek</span> | <span class ="tooltip" data-tooltip-text ="Field Value">8.005</span>     |
| 2015-08-18T00:06:00Z                                                            | below 3 feet                                                                        | santa_monica                                                              | 2.116                                                                    |
| 2015-08-18T00:12:00Z                                                            | between 6 and 9 feet                                                                | coyote_creek                                                              | 7.887                                                                    |
| 2015-08-18T00:12:00Z                                                            | below 3 feet                                                                        | santa_monica                                                              | 2.028                                                                    |
```

I'm actually going to make some CSS changes to make tables scrollable when they're wider than the screen width. If you'll make the change above ^, I'll merge the PR and then adjust the CSS.

SiavasFiroozbakht:
Very glad to help, this is done.

It's great to hear this change is already happening as it was an improvement I had in mind, but tried to keep the format consistent.

If you'd like me to help with the CSS as well after the page is updated, just let me know!

sanderson:
Looks great. I already made the CSS change. This will go up with the next docs build. Thanks @SiavasFiroozbakht!

PR #1794, opened 2018-08-29
Kapacitor/post handler update

sanderson:


content/kapacitor/v1.5/event_handlers/post.md
```diff
@@ -121,14 +175,14 @@ defined in the `kapacitor.conf`:
 _**httppost settings in kapacitor.conf**_  
 ```toml
 [[httppost]]
-  endpoint = "api/alert"
+  endpoint = "api-alert"
   url = "http://mydomain.com"
```
brettdh: When I originally read these docs, I had thought that `endpoint` would always be appended to `url`, because of its name. Could you please disambiguate this throughout these docs by always including a `/path` with every `url` and/or explicitly stating that `endpoint` is *just* an identifier and does not become part of the URL?

content/kapacitor/v1.5/event_handlers/post.md
```diff
@@ -296,7 +297,7 @@ _**kapacitor.conf**_
 ```toml
 [httppost]]
```
brettdh: Typo: unmatched brackets

content/kapacitor/v1.5/event_handlers/post.md
```diff
@@ -305,7 +306,7 @@ _**kapacitor.conf**_
...
   endpoint = "example"
-  url = "http://example.com"
+  url = "http://example.com/path"
```
brettdh: Typo: unmatched brackets

sanderson: Nice catch @brettdh. Fixed.

content/kapacitor/v1.5/event_handlers/post.md
```diff
@@ -29,7 +29,9 @@ Below is an example configuration:
 ```
 
 #### `endpoint`
-Name of a configured httppost endpoint. _**Can not be specified in the URL.**_
```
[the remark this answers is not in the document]
sanderson: This is a line that was removed/changed.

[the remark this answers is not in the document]
stevebang: oops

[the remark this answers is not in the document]
stevebang: I also wrote this, which applies to the entire page.

"HTTP POST endpoint" instead of "httppost." Later in this file too. Anywhere "post" refers to the HTTP POST request, it should be uppercased, including the title. Unfortunately, we're cleaning up what we got.

When the endpoint is referred to, should it be displayed as "/httppost" to make it clearer?

PR #1801, opened 2018-08-30
Kapactior enterprise alerts in clusters

sanderson:
Sorry, this is kind of a loaded PR. At its core, it's a guide for adding alerts to a Kapacitor cluster, but in the process of creating this guide, I also added a shortcode for including inline SVGs (used for the accompanying diagrams) and updated the CSS so Kapacitor Enterprise uses the Kapacitor theme and colors.

Here are screenshots of the included diagrams so you don't have to pull the PR and run the docs locally to see them:

image

image


sanderson:
These changes will actually be included in a larger set of changes related to Kapacitor Enterprise and cluster management. Closing this PR.

PR #1817, opened 2018-09-10
Added max-fetch note to InfluxDB Enterprise config doc, updated formatting

sanderson:


content/enterprise_influxdb/v1.6/administration/configuration.md
```diff
@@ -1014,52 +1010,56 @@ Note that increasing `retry-concurrency` also increases network traffic.
...
 
-### max-fetch = 10
+#### max-fetch = 10
 
 The maximum number of shards that a single data node will copy or repair in parallel.
 
 Environment variable: `INFLUXDB_ANTI_ENTROPY_MAX_FETCH`
 
-<br>
+> Having `max-fetch=10` with higher numbers of shards (100+) can add significant overhead to running nodes.
```
dgnorton: AE isn't currently using this config setting. It now defaults to 1 shard copy at a time per node. That is, each node could be copying one shard simultaneously. AE also defaults to 1 shard repair at a time per node. Copy is used to fix missing shards and repairs are used to fix existing shards that are missing data. By default, each node can be performing one of each type concurrently. As of 1.6, the only way to increase the number of concurrent repairs per node is by setting experimental environment variables. E.g.,
```
INFLUX_EXP_MAX_AE_COPY=3
INFLUX_EXP_MAX_AE_REPAIR=2
```
These are experimental and not yet intended for production.

sanderson: @timhallinflux with `max-fetch` not currently being used, should I just remove it from the doc?

timhallinflux: @dgnorton any issue in terms of removing this param from the docs?

sanderson:
I commented out the `max-fetch` AE setting with the expectation that it will return in a future version of InfluxDB enterprise.

PR #1819, opened 2018-09-10
Updated InfluxDB Prometheus read/write doc

sanderson:


content/influxdb/v1.6/supported_protocols/prometheus.md
```diff
@@ -8,40 +8,104 @@ menu:
...
+but you're welcome to use the whatever database name you like.
+
+```sql
+CREATE DATABASE "prometheus"
+```
+
 ### Configuration
 
-To enable the use of the Prometheus remote read and write API with InfluxDB, you need to add URL values to the following settings in the [Prometheus configuration file](https://prometheus.io/docs/prometheus/latest/configuration/configuration/):
+To enable the use of the Prometheus' remote read and write APIs with InfluxDB,
```
noramullen1: Is there an extra "the" here? ("To enable the use of the Prometheus'...)

PR #1835, opened 2018-09-19
Fixed formatting of Kapacitor Event Handler Setup guide

sanderson:


content/kapacitor/v1.5/guides/event-handler-setup.md
```diff
@@ -174,137 +178,163 @@ Configure Kapacitor to send alert messages to a Telegram bot.
...
+
+1. Search for the **@BotFather** username in your Telegram application.
+
+2. Click **Start** to begin a conversation with **@BotFather**.
+
+3. Send `/newbot` to **@BotFather**.
+
+    **@BotFather** responds:
+
+    _Alright, a new bot. How are we going to call it? Please choose a name for your bot._
```
noramullen1: Should be 'what' instead of 'how' here.

content/kapacitor/v1.5/guides/event-handler-setup.md
```diff
@@ -174,137 +178,163 @@ Configure Kapacitor to send alert messages to a Telegram bot.
...
-Your newly-created bot will appear in the chat list on the left side of the application.
->
 
-##### Telegram API access token
+6. Begin a conversation with your bot.
+
+    Click on the `t.me/<bot-username>` link in **@BotFather**'s response
+    and click **Start** at the bottom of your Telegram application.
+
+    Your newly-created bot will appear in the chat list on the left side of the application.
```
noramullen1: No hyphens after words that end in -ly.

content/kapacitor/v1.5/guides/event-handler-setup.md
```diff
@@ -174,137 +178,163 @@ Configure Kapacitor to send alert messages to a Telegram bot.
...
-`global`
-&emsp;&emsp;&emsp;Set to `true` to send all alerts to Telegram without needing to specify Telegram in TICKscripts.
-`state-changes-only`
-&emsp;&emsp;&emsp;Set to `true` to only send an alert to Telegram if the alert state changes. This setting only applies if the `global` setting is also set to `true`.
+`chat_id`  
+Set to your Telegram [chat ID](#telegram-chat-id). This serves as the default chat ID if the TICKscript doesn't specify a chat ID.
 
-#### Sample configuration
+`parse-mode`  
+Set to `Markdown` or `HTML` for markdown-formatted or HTML-formatted alert messages.
```
noramullen1: Capital M for Markdown

content/kapacitor/v1.5/guides/event-handler-setup.md
sanderson: This is the response that comes from Telegram's bot and is the actual verbiage they use. I think "How" is used because referencing a bot is often referred to as "calling" a bot. It's referring to the behavior of referencing the bot rather than naming it. Confusing, I know...

noramullen1:
Ah got it. I think it's an ESL issue...either way, agree that you should
use their language.

PR #1862, opened 2018-10-04
Changed the way tabbed content is rendered to be more mobile friendly

sanderson:
There vertical tabs used across the docs were anything but mobile friendly, so I updated them to scale better. Here's some screenshots:

_**Desktop**_
image

_**Mobile**_
image


stevebang:
When tabs are used, the usual practice seems to be that the background is coordinated with the tab color.  But given a black background, I'm not sure how this would be done.  I think what you've done is an improvement -- we lose less space on the left and the tab contents being highlighted show what is covered below them.

For what I mean about tabs, see this: https://www.nngroup.com/articles/tabs-used-right/

PR #1874, opened 2018-10-10
Flux introduction doc

sanderson:
This is the very beginning of a user's journey into what Flux is and starts them on the path of writing a Flux script. The next few docs in the chain will walk through what a bucket is, how to pull data from a source, how to filter, range, etc.


sanderson:
I was planning on introducing this in the getting started guide, but I don't think it would be a bad thing to introduce here as well.

PR #1875, opened 2018-10-10
Note cell type for 1.7

noramullen1:


.gitignore
```diff
@@ -5,3 +5,4 @@ deploy
 .*.swp
 node_modules
 *.log
+content/platform/introduction/getting-started.md
```
sanderson: @Nora, this shouldn't be added to the `.gitignore`. This will prevent it from being committed/pushed/pulled.

PR #1878, opened 2018-10-11
Platform troubleshooting oom loop doc

sanderson:


content/platform/troubleshooting/oom-loops.md
```diff
@@ -1,11 +1,76 @@
...
+restarting, it continues to use more and more RAM until the cycle repeats.
+
+In a [monitoring dashboard](/platform/monitoring/monitoring-dashboards), an OOM loop
+will appear in the **Memory Usage %** metric and look similar to the following:
+
+![OOM Loop](/img/platform/troubleshooting-oom-loop.png)
+
+## Potential causes
+The causes of OOM loops vary widely and depend on your specific use case of the TICK stack.
+The following are common causes:
```
stevebang: Only one cause is mentioned -- either reword or add more causes.

sanderson: I'll reword it. Been waiting on feedback and other potential causes from the support team but haven't heard anything back yet.

content/platform/troubleshooting/oom-loops.md
```diff
@@ -1,11 +1,76 @@
...
+    5 root       0 -20       0      0      0 S   0.0  0.0   0:00.00 kworker/0:0H
+    6 root      20   0       0      0      0 S   0.0  0.0  12:29.14 kworker/u30:0
+    7 root      20   0       0      0      0 R   0.0  0.0 186:58.33 rcu_sched
+    8 root      20   0       0      0      0 S   0.0  0.0 546:06.54 rcuos/0
+    9 root      20   0       0      0      0 S   0.0  0.0 547:13.28 rcuos/1
+   10 root      20   0       0      0      0 S   0.0  0.0   0:00.00 rcuos/2
+   11 root      20   0       0      0      0 S   0.0  0.0   0:00.00 rcuos/3
+```
+
+> In order to correctly identify the culprit, this command should be run while memory usage is high.
```
noramullen1: Maybe reword second half of this to avoid passive voice ("run this command while..." instead of "this command should be run").

noramullen1: Or "we recommend running this..."

content/platform/troubleshooting/oom-loops.md
```diff
@@ -1,11 +1,155 @@
...
+    9 root      20   0       0      0      0 S   0.0  0.0 547:13.28 rcuos/1
+   10 root      20   0       0      0      0 S   0.0  0.0   0:00.00 rcuos/2
+   11 root      20   0       0      0      0 S   0.0  0.0   0:00.00 rcuos/3
+```
+
+> In order to correctly identify the culprit, run this command while memory usage is high.
+
+<!-- -->
+
+> `top` and other process monitoring tools vary by operating system.
```
stevebang: Is this a note? Needs a "Note:" label?

sanderson: This is a note. I feel like prepending a styled notice with "Note:" is redundant. In the rendered and styled HTML, it's already set apart as something important (in the green box).

timhallinflux: We know the process.....so I don't believe this is necessary.  The process is `influxd` or `influd-data`.

An OOM loop will occur as a result of a repeated issuance of a query which exhausts memory.  I.e. a dashboard with a new cell....which is set on refresh (every 30s).  

One way to check on the query side of this is to attempt to look at the logs to see if such a query is being issued --- you can use SHOW QUERIES -- because the process is dying and restarting.

The fix for an OOM loop is "easy" if you are running in a virtualized environment or a cloud environment where you can easily change the resources available to the process -- in this case allocating more RAM.  In environments where you are running on a fixed set of resources, this can be a difficult challenge to overcome.

sanderson: Removed the `top` instructions. Added information about viewing logs and scaling the available memory.

PR #1903, opened 2018-10-19
Fix/incorrect description

### LINKED ISSUE #1841: Typo issue
According to docs in, https://docs.influxdata.com/influxdb/v1.6/concepts/key_concepts/, `tags`  store the meta data and `fields` store the measured value.

In the immediate doc `key_concepts`, https://docs.influxdata.com/influxdb/v1.6/concepts/key_concepts/, its given that `field keys` store metadata. (Sample Data -> Discussion -> Para 3 -> Line 2).

Correct me, in case of an invalid issue report/misunderstanding. TIA

sdomino:
Fixed incorrect description of 'field keys' (Resolves #1841)


sdomino:
Cleaned up in #1904

PR #1904, opened 2018-10-19
Fixed incorrect description (Resolves #1841)

### LINKED ISSUE #1841: Typo issue

sdomino:
Was previously #1903, but for some reason, that one got all muddied up with a bunch of commits.


sdomino:
Formerly #1903, but now cleaned up commits.

sanderson:
Thanks @sdomino!

PR #1915, opened 2018-10-24
Fix Enterprise labels and add remaining missing descriptions

stevebang:


content/platform/monitoring/tools/measurements-internal.md
```diff
@@ -13,12 +13,12 @@ menu:
 * [Using the `_internal` database](#using-the-internal-database)
 * [Enterprise cluster measurements](#enterprise-cluster-measurements)
 * [InfluxDB measurement statistics](#influxdb-measurement-statistics)
-  * [ae](#ae-e) (E)
+  * [ae](#ae-e) (Enterprise only)
```
sanderson: The link anchor is wrong here. It needs to be switched to `(#ae-enterprise-only)`.

stevebang: Fixed.

stevebang: Fixed links here and in Enterprise Cluster Stats monitoring dashboard.

PR #1967, opened 2018-11-07
influxdb: Update influxdb default Unix socket path

guillemj:
This matches the change done in influxdb.

The PR is on its way to the influxdb repo. I'm also not sure how you handle updates for projects' master (f.ex. influxdb or telegraf) as there's no such unversioned directory, so you might want to have one so that these can be staged and then copied into the next released version, or change the PR template in influxdb, as it's recommending to send these doc PRs. :)


guillemj:
Ok, another empty reply just to undo the stale annoyance… this is still relevant.

pierwill:
https://github.com/influxdata/influxdb/pull/10462 closed.

PR #1970, opened 2018-11-07
Flux histograms guide

sanderson:


sanderson:
@timhallinflux I agree with the potential overuse of "buckets." Since it's the actual parameter name, are we good to use it in this doc or should be broach the subject of a potential parameter rename with the Flux team?

timhallinflux:
I think we go with this "as-is" -- and yes, we should broach the subject with the Flux team.  It's good hygiene.

sanderson:
One challenge is that "buckets" is a term commonly used to refer to data segments when creating histograms. The only other option I've found commonly used is "bins."

aanthony1243:
Changing parameter names is not a burden on the `flux` team.  we want to get it right as much as you do.  wikipedia uses the term `bin` first, and `bucket` second.  That's not a definitive answer but good enough for me.  I'd guess most people who know what a histogram is will be just fine seeing `bin` as the parameter name.  

less standard alternatives: segments, cuts, or sections?

aanthony1243:
if it's front-of-mind for you, it'd be great if you can give us a check-list of names that need to change and then we can sync the change on a scheduled release with docs.

PR #1986, opened 2018-11-13
Fix formatting and add clarification re. new dashboards

noramullen1:


noramullen1:
@timhallinflux Updated with those suggestions.

timhallinflux:
@russorat -- have a peek at this and make sure you are happy with this treatment.  It's tough teasing apart the host specific, pre-canned, and unchangable "dashboards" from the proto-board, dashboard templates which can span hosts are changable ...etc.   They appear in different spots...etc.

content/chronograf/v1.7/guides/using-precreated-dashboards.md
```diff
@@ -11,9 +11,9 @@ menu:
 
 ## Overview
 
-Pre-created dashboards are delivered with Chronograf and are available depending on which Telegraf input plugins you have enabled. These dashboards include cells with data visualizations for metrics that are relevant to data sources you are likely to be using. The JSON files that are delivered include headings, queries, labels, and other information that can also be useful for helping you get started in [creating custom dashboards](/chronograf/latest/guides/create-a-dashboard/).
+Pre-created dashboards are delivered with Chronograf and are available depending on which Telegraf input plugins you have enabled. These dashboards include cells with data visualizations for metrics that are relevant to data sources you are likely to be using.
```
russorat: IMO, we should mention here that these pre-created dashboards are only a subset of the host list page instead of part of the note below. We could even go so far as to call them "Host Detail Views" or something similar and not call them dashboards at all.

PR #2002, opened 2018-11-16
Clarified 'telegraf  database name' field on wizard

noramullen1:


content/chronograf/v1.7/administration/creating-connections.md
```diff
@@ -27,7 +27,7 @@ To create an InfluxDB connection in the Chronograf UI:
       *Required only if [authorization is enabled](/influxdb/latest/administration/authentication_and_authorization/) on the InfluxDB instance to which you're connecting.*
     * **Password**: Enter the password.
       *Required only if [authorization is enabled](/influxdb/latest/administration/authentication_and_authorization/) on the InfluxDB instance to which you're connecting.*
-    * **Telegraf Database Name*: Enter the name of the Telegraf database. The field is prefilled with `telegraf`.
+    * **Telegraf Database Name**: Enter the name of your Telegraf database. This field specifies the database that Chronograf will use to for plugins that populate the Host List page. The field is prefilled with `telegraf`, which is the default name assigned to Telegraf databases. If you've specified a different name for your Telegraf database, enter that here instead.
```
russorat: maybe something like this:
>This field specifies the database that Chronograf will use for populating different parts of the application, including the host list page. You will still be able to query any database you have access to in the InfluxDB instance when building dashboards or exploring data.

The second line comes from some feedback during InfluxDays where people thought that this was the only database in the source they could query from.

content/chronograf/v1.7/administration/creating-connections.md
```diff
@@ -27,7 +27,7 @@ To create an InfluxDB connection in the Chronograf UI:
...
     * **Default Retention Policy**: Enter the name of the default [retention policy](/influxdb/latest/concepts/glossary/#retention-policy-rp). If none i s provided, it assumes `autogen`.
```
russorat: How about something like: 
> This is the name of the default retention policy Chronograf will use when one is not specified. If you've changed the default retention policy in your InfluxDB instance, you may want to change it here as well. If none is provided, we will use `autogen`.

noramullen1:
Thanks Russ! Updated with your suggestions.

PR #2015, opened 2018-11-27
Flux operators

sanderson:


content/flux/v0.7/functions/custom-functions.md
```diff
@@ -54,7 +54,7 @@ multiply = (x, y) => x * y
 ## Functions that manipulate pipe-forwarded data
 Most Flux functions manipulate data pipe-forwarded into the function.
```
aanthony1243: not sure if this is something we want to introduce, since it may confuse the users, but pipe-forwarding is not exclusive to tables.  You can write a function to take a pipe-forward value of any type.  What you do have to ensure is that the output of the first function is acceptable for the pipe-forward argument of the second function. 

While it's possible, for example, to pipe-forward an integer or string, I don't think we have an actual compelling example of this use-case.

content/flux/v0.7/language/operators.md
```diff
@@ -0,0 +1,98 @@
...
+## Arithmetic operators
+Arithmetic operators take two numerical values (either literals or variables) and
+perform a calculation that returns a single numerical value.
+
+| Operator | Description    | Example  | Result |
+|:--------:| -----------    | -------  | ------ |
+| `+`      | Addition       | `1 + 1`  | `2`    |
+| `-`      | Subtraction    | `3 - 2`  | `1`    |
+| `*`      | Multiplication | `2 * 3`  | `6`    |
+| `/`      | Division       | `9 / 3`  | `3`    |
```
aanthony1243: it will be important to clarify when results are integers or floats.  I have to admit, it's not immediately known to me what the behavior is here.  IIRC, the output depends on the inputs.  if we have 2 ints  ( 9 / 4 ) you get integer division (2).  But if you have one or both arguments as floats  (9/4.0)  you get a floating point result (2.25).

sanderson: Added a note about not currently not being able to mix numeric types in arithmetic operations. This is related to https://github.com/influxdata/flux/issues/476.

content/flux/v0.7/language/operators.md
```diff
@@ -0,0 +1,102 @@
...
+## Assignment operators
+An assignment operator assigns a value to its left operand based on the value of its right operand.
+
+| Operator | Description                                         | Example | Meaning |
+|:--------:| -----------                                         | ------- | ------- |
+| `=`      | Assign value of left expression to right expression | `x = y` | x = y   |
+
+
+## Function operators
+Function operators are used to create functions and control the flow of data through function operations.
```
noramullen1: Can we cut "are used to"?

content/flux/v0.7/language/operators.md
```diff
@@ -0,0 +1,102 @@
...
+| `!=`     | Not equal to                    | `"abc" != "def"`    | `true`  |
+| `<`      | Less than                       | `1 < 2`             | `true`  |
+| `>`      | Greater than                    | `1 > 2`             | `false` |
+| `<=`     | Less than or equal              | `1 <= 2`            | `true`  |
+| `>=`     | Greater than or equal           | `1 >= 2`            | `false` |
+| `=~`     | Equal to regular expression     | `"abc" =~ /[a-z]*/` | `true`  |
+| `!~`     | Not equal to regular expression | `"abc" !~ /[0-9]*/` | `true`  |
+
+
+> The `>` and `<` operators can be used to [compare the lexicographic order of strings](#string-operators).
```
noramullen1: Can we cut "can be used to" here?

content/flux/v0.7/language/operators.md
```diff
@@ -0,0 +1,102 @@
...
+String operators concatenate or compare string values.
+
+| Operator | Description                         | Examples        | Result  |
+|:--------:| -----------                         | --------        | ------  |
+| `+`      | Concatenation                       | `"ab" + "c"`    | `"abc"` |
+| `<`      | Less than in lexicographic order    | `"ant" < "bee"` | `true`  |
+| `>`      | Greater than in lexicographic order | `"ant" > "bee"` | `false` |
+
+## Literal constructors
+Literal constructors are used to define fixed values.
```
noramullen1: Cut "are used to"?

PR #2019, opened 2018-12-03
Kapacitor scheduled downtime guide

sanderson:


content/kapacitor/v1.5/guides/scheduled-downtime.md
```diff
@@ -0,0 +1,196 @@
...
+In many cases, infrastructure downtime is necessary to perform system maintenance.
+This type of downtime is typically scheduled beforehand, but can trigger unnecessary
+alerts if the affected hosts are monitored by Kapacitor.
+This guide walks through creating TICKscripts that gracefully handle scheduled downtime
+without triggering alerts.
+
+## Sideload
+Avoiding unnecessary alerts during scheduled downtime is accomplished by using
+the `sideload` node to load information from files in the filesystem
+and set fields and tags on data points which can then be used in alert logic.
```
noramullen1: To avoid passive voice: "Avoid unnecessary alerts during schedule downtime by using...."

content/kapacitor/v1.5/guides/scheduled-downtime.md
```diff
@@ -0,0 +1,196 @@
...
+        └── hostgroups/
+            ├── cluster7.yml
+            └── cluster8.yml
+```
+
+> You only need to create files for hosts or hostgroups that will be offline.
+
+The contents of the file should contain one or more key-value pairs.
+The key being the field or tag key that will be set on each matching point,
+and the value being the field or tag value that will be set on matching points.
```
noramullen1: This is a fragment.

content/kapacitor/v1.5/guides/scheduled-downtime.md
```diff
@@ -0,0 +1,196 @@
...
+Kapacitor will iterate through the different values for each tag and search for
+matching files in the source directory.
+
+```js
+|sideload()
+  .source('file:///usr/kapacitor/scheduled-maintenance')
+  .order('hosts/{{.host}}.yml' , 'hostgroups/{{.hostgroup}}.yml')
+```
+
+The order in which file path templates are listed in the `order` property define
```
noramullen1: Can we say "The order of file path templates in the 'order' property..."? Lots of instances of passive voice in this are tricky/impossible to write around so just trying to clean up the few ones that are easier.

content/kapacitor/v1.5/guides/scheduled-downtime.md
```diff
@@ -0,0 +1,196 @@
...
+the precedence in which file paths are checked.
+Those listed first, from left to right, are checked first.
+
+### Define the sideload field
+The `field` property requires two arguments:
+
+1.  The key that Kapacitor looks for in the source files and the field for which
+    it defines a value on each data point.
+2.  The default value used if no matching file and key are found in the source files.
```
noramullen1: These aren't sequential steps, so they should be bullet points instead of numbered.

content/kapacitor/v1.5/guides/scheduled-downtime.md
```diff
@@ -0,0 +1,196 @@
...
+This assumes hosts are not undergoing maintenance by default.
+
+```js
+|sideload()
+  .source('file:///usr/kapacitor/scheduled-maintenance')
+  .order('hosts/{{.host}}.yml' , 'hostgroups/{{.hostgroup}}.yml')
+  .field('maintenance', FALSE)
+```
+
+> The `tag` property could be used instead of `field` if you prefer to set a tag
```
noramullen1: "You can use the 'tag' property instead of 'field'...."

content/kapacitor/v1.5/guides/scheduled-downtime.md
```diff
@@ -0,0 +1,196 @@
...
+> on each data point rather than a field.
+
+### Update alert logic
+The `sideload` node will now set the `maintenance` field on every data point processed by the TICKscript.
+For those that have `host` or `hostgroup` tags matching the filenames of the source files,
+the `maintenance` field will be set to the value define in the source file.
```
noramullen1: Missing D at end of "defined"

content/kapacitor/v1.5/guides/scheduled-downtime.md
sanderson: They are sequential in this context. They are positional arguments. Kapacitor assumes the first argument is the key and the second argument is default value.

I did end up restructuring the content a little to be more clear.

content/kapacitor/v1.5/guides/scheduled-downtime.md
```diff
@@ -0,0 +1,205 @@
...
+from various file-based sources.
+
+Kapacitor searches the specified files for a given field or tag key.
+If it finds the field or tag key in the loaded files, it uses the value in the files to
+set the field or tag on data points.
+If it doesn't find the field or tag key, it sets them to the default value defined
+in the [`field` or `tag` properties](#field).
+
+### Relevant sideload properties
+Of `sideload`'s properties, the following are relevant to gracefully handling scheduled downtime:
```
stevebang: The following `sideload` properties are ... 
or
The following properties of `sideload` are ...

PR #2020, opened 2018-12-04
Flux - updated group function

sanderson:
This change will likely be included with Flux v0.7.2.


sanderson:
No additional docs will be released for Flux's v0.7 docs. The next public release will be a newer minor version, therefore these changes should note be in the `flux/v0.7` directory.

These changes are included in https://github.com/influxdata/docs.influxdata.com/commit/93418b3e8015c8c383c722bedbb15eed872514f7 which is part of a future release branch for Flux docs.

PR #2051, opened 2018-12-06
Remove fromRows, add influxFieldsAsCols

sanderson:


sanderson:
No additional docs will be released for Flux's v0.7 docs. The next public release will be a newer minor version, therefore these changes should note be in the `flux/v0.7` directory.

These changes are included in https://github.com/influxdata/docs.influxdata.com/commit/3e8923ac8a8c62bb121eb59bf351c908755bf499 which is part of a future release branch for Flux docs.

PR #2069, opened 2018-12-14
Flux – assertEquals function

sanderson:


content/flux/v0.x/functions/tests/_index.md
```diff
@@ -0,0 +1,14 @@
+---
+title: Flux testing functions
+description: Flux testing functions test piped forward data in specific ways and return errors if the tests fail.
```
noramullen1: Hyphenate "piped-forward data" here like below, otherwise it's a bit confusing.

PR #2083, opened 2018-12-29
Replace curl with wget in Linux context

hongquan:
On Linux, `wget` is more ubiquitous than `curl`. Most distros have `wget` pre-included, while `curl` needs to be installed afterwards.

So, it is better that the documentation tell user to use `wget` to download InfluxDB. They just copy the sample command and run, don't have to check if `curl` exist and install it first.

Other parts of documentation where `curl` is mentioned, I don't replace with `wget` because it seems not to be specific to Linux. Those commands seem to target both Linux and MacOS.

### DIFF
--- a/content/influxdb/v0.10/introduction/installation.md
+++ b/content/influxdb/v0.10/introduction/installation.md
@@ -35,7 +35,7 @@ Debian and Ubuntu users can install the latest stable version of InfluxDB using 
 For Ubuntu users, you can add the InfluxData repository by using the following commands:
 
 ```bash
-curl -sL https://repos.influxdata.com/influxdb.key | sudo apt-key add -
+wget -qO- https://repos.influxdata.com/influxdb.key | sudo apt-key add -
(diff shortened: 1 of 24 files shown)


sanderson:
Thanks @hongquan! In order for me to approve the contribution, you must first sign the [InfluxData Contributor License Agreement (CLA)[https://www.influxdata.com/legal/cla/].

I think this is a good change. Another option is to include both curl and wget commands. I know alot of people have strong opinions about which to use.

sanderson:
Thanks for submitting this @hongquan. I added some functionality that allows you to define "code tabs" and switch between variations of code blocks. I've used your proposed changes in that PR, although I didn't do it for all versions of InfluxDB and Telegraf; only the latest two versions.

Thank you again for your contribution. I'm closing this PR in favor of https://github.com/influxdata/docs.influxdata.com/pull/2135, we really appreciate the suggestion and the modified commands.

PR #2089, opened 2019-01-04
Update managing-security.md

hydrandt:
From the github interface it might seem like the oauth app can see all your organizations memberships. Adding an explicit note about the necessity to grant the organization access, so noone gets confused like I did in https://github.com/influxdata/chronograf/issues/4575


sanderson:
Thanks @hydrandt! In order for me to approve the contribution, you must first sign the [InfluxData Contributor License Agreement (CLA)[https://www.influxdata.com/legal/cla/].

The change looks good. My only suggestion would be to format this as a blockquote which gets styled as a green note. This information is important enough that it should be separated out from the other content.

hydrandt:
CLA signed, and good call - note formatted as blockquote.

PR #2094, opened 2019-01-05
removed background on dark theme, resolves #1943

### LINKED ISSUE #1943: Scrolling the docs is very slow on Ubuntu - Chromium and UP/DOWN is not working on keyboard.
The docs pages are very slow | unresponsive to scroll on Ubuntu 16, newest Chromium browser.

* Slow using the mouse wheel.
* Slow using the UP/DOWN arrows on a keyboard.
* It is NOT as slow on Firefox.

Related issue:
* When clicking  'Getting Started' on the left side, you cannot anymore use UP/DOWN on keyboard to scroll the page, which is annoying. You have to click the center page again to move the focus to is.
* This happens both on Chromium and Firefox
https://docs.influxdata.com/influxdb/v1.6/introduction/getting-started/

###### URL for relevant page?
https://docs.influxdata.com/influxdb/v1.6/introduction/getting-started/

Ubuntu 16 + Latest Chromium and Firefox
This has been an issue for ~ a year but now I was bothered enough to report it.

I looked for 'slow' in the issues.

tsaarni:
Like originally reported in #1943: scrolling the documentation is slow on Chrome on Linux.  The problem is caused by the SVG background image in the dark theme.  

This pull request removes the background image.  I hope this is acceptable solution, since the visible difference is very minor. Background is only used in "pre" tags and it is hardly noticeable.


tsaarni:
I have now submitted the CLA.  You might have it twice, I did the same last week according to [CONTRIBUTING.md](https://github.com/influxdata/docs.influxdata.com/blob/master/CONTRIBUTING.md) which has bit different URL, so not 100% sure if you got it too.

And by the way, it is of course ok to me if you'd rather not merge this directly but instead re-generate static content with gulp yourself (no need to trust the files I generated!) :)

I also realized that original ticket had also second issue, related to focus, which is NOT solved by this PR!

sanderson:
You certainly did sign the CLA before. Sorry, when I searched it, I must have typed your GH username in wrong the first time. I'm fine with merging this directly in. The generated css files look fine. Thanks again @tsaarni !

PR #2151, opened 2019-02-15
Fix description for subscription-mode "cluster"

yousong:
As of kapacitor OSS 1.5.2, "kapacitord config" output indicates that subscription-mode defaults to "cluster" mode.

NOTE: please consider also adding docs to the shipped configuration `/etc/kapacitor/kapacitor.conf`.   Currently the comment there (`Subscription mode is either "cluster" or "server"`) does not give much information on the actual effect of each option value.

### DIFF
--- a/content/kapacitor/v1.5/administration/subscription-management.md
+++ b/content/kapacitor/v1.5/administration/subscription-management.md
@@ -74,12 +74,7 @@ Defines the subscription mode of Kapacitor.
 Available options:
 
 - `"server"`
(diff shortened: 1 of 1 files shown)


sanderson:
Thanks @yousong for pointing this out. The warning around using `cluster` was wrong, but a warning still needs to be there. I've corrected the message in https://github.com/influxdata/docs.influxdata.com/pull/2152 and am closing this PR in favor of that one. Thank you very much for submitting the suggestion though.

PR #2156, opened 2019-02-20
kapacitor: Add unit info on fields added by state tracking node

yousong:
The description text for stateDuration node seems to imply that the
additional field value will be of duration type while actually it's
float64 describing number of .unit()

While at it, also convert the paragraph to "semantic linefeeds" style


sanderson:
Thanks @yousong!

Issue #2166, opened 2019-02-26
Upgrade instructions -- improve clarity on "remove from load balancer" step


timhallinflux:
###### URL for relevant page?
https://docs.influxdata.com/enterprise_influxdb/v1.7/administration/upgrading/
and previous editions of upgrade instructions which include instructions about the "load balancer"

###### What products and version are you using?
InfluxDB Enterprise

###### Where did you look before opening the issue?
Two customers have raised support cases regarding the upgrade instructions which indicate some confusion with this particular step.

https://docs.influxdata.com/enterprise_influxdb/v1.7/administration/upgrading/#step-2-remove-the-data-node-from-the-load-balancer

First, what is happening here?
In the production installation instructions, we indicate:
https://docs.influxdata.com/enterprise_influxdb/v1.7/install-and-deploy/production_installation/data_node_installation/#load-balancer

That a load balancer should be used to distribute read and write traffic across available data nodes.  However, when a data node is taken down for maintenance --- the customer can proactively remove the node from the load balancer PRIOR to taking the node offline.  

Second, What does this do?
Effectively, removing the data node from the load balancer's configuration prevents the load balancer from distributing traffic to a node that is offline.  What typically happens is that if the customer DOES NOT remove the data node from the LB configuration prior to taking it offline, the load balancer will attempt to route traffic to the dead node.  Since the node is unavailable, the load balancer (depending on it's features and functions) may attempt to re-route the traffic to other available nodes.  Or, if may simply report a failure back to the requestor (read or write).  Typically, load balancers will STOP routing traffic to nodes after some number of predefined failures (3-5 is typical), so, this tends to be a temporary issue.  Once the node returns to a healthy state, LBs may also re-attempt to route to it (based on a "health check" or ping that the LB performs against the node in the routing table). Again, this is based on the underlying features, capabilities and configuration of the LB itself. So, the specifics of configuring all of this is outside of the scope of our documentation. 

So, the best practice is to change the load balancer configuration to avoid triggering any failures back to the requestors.  Is this absolutely required, no.  

There will be customers who have access to the software configuration "stack" -- but may NOT have access to the networking "stack" (i.e. LBs).  In those cases, they may choose to simply have the ability  (without significant effort or coordination with other teams) to change the LB.  So, they may choose to simply endure a small number of failures while they upgrade the database.

kelseiv:
resolved in #2168

PR #2168, opened 2019-02-27
update Step 2 & 7 per #2166

kelseiv:


kelseiv:
# 8322676c4a5b9c4ef56994137af0b6ce328ddcc9 
addresses first part of #2166 
The second part of #2166 is:
Adding best practices docs for load balancer, including potential issues that may arise when traffic continues to a data node during an upgrade. And then, referring this doc in upgrade instructions....the best place for this content is TBD.

content/enterprise_influxdb/v1.7/administration/upgrading.md
```diff
@@ -115,10 +115,11 @@ wget https://dl.influxdata.com/enterprise/releases/influxdb-data_1.7.4-c1.7.4_am
 wget https://dl.influxdata.com/enterprise/releases/influxdb-data-1.7.4_c1.7.4.x86_64.rpm
 ```
 
-### Step 2: Remove the data node from the load balancer
+### Step 2: Change load balancer configuration
```
sanderson: I think this heading may be too generic. The important concept to understand in this section is that you need to stop read/write traffic to the data node you're upgrading. This is done by updating the load balancer. I think a more specific title could be:

> Stop traffic to the data node

This would pair well with the updated "Step 7" heading below. The specific instructions involving the load-balancer could then be included in the next paragraph.

content/enterprise_influxdb/v1.7/administration/upgrading.md
```diff
@@ -115,10 +115,11 @@ wget https://dl.influxdata.com/enterprise/releases/influxdb-data_1.7.4-c1.7.4_am
...
 
-To avoid downtime and allow for a smooth transition, remove the data node you are upgrading from your
-load balancer **before** performing the remaining steps.
+If you have access to the load balancer configuration, stop routing read/write requests to the data node server via your load balancer (port 8086) **before** performing the remaining steps.
```
sanderson: Maybe mention what a load balancer is before this sentence. For example:

> Your cluster's load-balancer distributes read and write requests among data nodes in the cluster.

content/enterprise_influxdb/v1.7/administration/upgrading.md
```diff
@@ -207,11 +208,11 @@ service influxdb restart
 sudo systemctl restart influxdb
 ```
 
-### Step 7: Add the data node back into the load balancer.
+### Step 7: Restart traffic to data node 
 
-Add the data node back into the load balancer to allow it to serve reads and writes.
+Restart routing read/write requests to the data node server via your load balancer (port 8086).
```
sanderson: Port 8086 is the port on which the load-balancer forwards read/write requests to InfluxDB. I think it me be clearer if  "port 8086" was right after "data node server":

> Restart routing read/write requests to the data node server on port 8086 via your load balancer.

PR #2170, opened 2019-02-28
Update installation.md

YaguraStation:
Add missing dep `apt-transport-https`which is not installed by default in Debian.


YaguraStation:
Nah, then I'd opt for leave-as-is without making the instructions bulkier. Minimal install users will figure this out by themselves and _apt_ has comprehensive feedback.

Issue #2173, opened 2019-03-04
Typo in "Kafka event handler" doc page.


prabathabey:
###### URL for relevant page?
https://docs.influxdata.com/kapacitor/v1.5/event_handlers/kafka/

The line in the page that reads "Create a handler file that subscribes to the cpu topic and uses the Kafka event handler to send alerts to the **alerts** channel in Kafka.." refers to publishing alerts to a Kafka topic called "alerts". However, the sample handler configuration "kafka_cpu_handler.yaml" says the alerts are published to a topic called "cpu-alerts".

PR #2174, opened 2019-03-04
Fixed typo in kafka event handler page.

prabathabey:
Fixes the issue reported in https://github.com/influxdata/docs.influxdata.com/issues/2173


sanderson:
Thanks @prabathabey!

PR #2199, opened 2019-03-25
add "configure gzip compression"

kelseiv:
Hi Daniel, pls let me know if you have suggestions/edits. Thx!

### DIFF
--- a/content/influxdb/v1.7/guides/writing_data.md
+++ b/content/influxdb/v1.7/guides/writing_data.md
@@ -7,22 +7,21 @@ menu:
     parent: Guides
 ---
 
-There are many ways to write data into InfluxDB including the [command line interface](/influxdb/v1.7/tools/shell/), [client libraries](/influxdb/v1.7/clients/api/) and plugins for common data formats such as [Graphite](/influxdb/v1.7/write_protocols/graphite/).
-Here we'll show you how to create a database and write data to it using the built-in HTTP API.
+Write data into InfluxDB using the [command line interface](/influxdb/v1.7/tools/shell/), [client libraries](/influxdb/v1.7/clients/api/), and plugins for common data formats such as [Graphite](/influxdb/v1.7/write_protocols/graphite/). (can all of these methods use the HTTP API?)
 
 > **Note**: The following examples use `curl`, a command line tool that transfers data using URLs. Learn the basics of `curl` with the [HTTP Scripting Guide](https://curl.haxx.se/docs/httpscripting.html).
 
-## Creating a database using the HTTP API
+### Create a database using the HTTP API
 To create a database send a `POST` request to the `/query` endpoint and set the URL parameter `q` to `CREATE DATABASE <new_database_name>`.
-The example below sends a request to InfluxDB running on `localhost` and creates the database `mydb`:
-<br>
+The example below sends a request to InfluxDB running on `localhost` and creates the `mydb` database:
 
 ```bash
 curl -i -XPOST http://localhost:8086/query --data-urlencode "q=CREATE DATABASE mydb"
 ```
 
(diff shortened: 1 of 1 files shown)


content/influxdb/v1.7/guides/writing_data.md
```diff
@@ -44,6 +43,29 @@ The timestamp - supplied at the end of the line in Unix time in nanoseconds sinc
 If you do not specify a timestamp InfluxDB uses the server's local nanosecond timestamp in Unix epoch.
 Anything that has to do with time in InfluxDB is always UTC.
 
+### Configure gzip compression
+
+InfluxData supports gzip compression. To increase data transfer speed, consider the following options:
```
danielnelson: I think you mean InfluxDB not InfluxData.

> increase data transfer speed

Data transfer speed is the same, I would say  "reduce network traffic".

content/influxdb/v1.7/guides/writing_data.md
```diff
@@ -44,6 +43,29 @@ The timestamp - supplied at the end of the line in Unix time in nanoseconds sinc
...
+
+InfluxData supports gzip compression. To increase data transfer speed, consider the following options: 
+
+  * To accept compressed data from InfluxDB, add the `Accept-Encoding: gzip` header to HTTP API requests.
+
+  * To compress data before sending it to InfluxDB, add the `Content-Encoding: gzip` header to HTTP API requests.
+
+  For details about enabling gzip for client libraries, see your client library documentation. 
+  
+  To enable gzip for Telegraf, add the `Content_encoding = “gzip”` header to your Telegraf configuration file (telegraf.conf).
```
danielnelson: > To enable gzip for Telegraf

This setting enables gzip encoding in the influxdb or influxdb_v2 outputs only.

The Telegraf option is lowercase: `content_encoding = "gzip"`.  Don't use fancy quotes here as it will cause problems.

Header is also not the right word here, I think it will be confusing, and this doesn't describe well enough where to put the setting in the config.  IMHO this is a place where an example would be useful.

content/influxdb/v1.7/guides/writing_data.md
```diff
@@ -44,6 +43,29 @@ The timestamp - supplied at the end of the line in Unix time in nanoseconds sinc
...
+
+  * To accept compressed data from InfluxDB, add the `Accept-Encoding: gzip` header to HTTP API requests.
+
+  * To compress data before sending it to InfluxDB, add the `Content-Encoding: gzip` header to HTTP API requests.
+
+  For details about enabling gzip for client libraries, see your client library documentation. 
+  
+  To enable gzip for Telegraf, add the `Content_encoding = “gzip”` header to your Telegraf configuration file (telegraf.conf).  
+
+  **gzip compression considerations**
```
danielnelson: I would remove this section, I don't think it is helpful.

> Determine the appropriate compression level: 1 (fastest) - 9 (most compressed).

Nobody worries about this, they just use the defaults.  Anyone who wants to change this will already know about it.

> Identify the types of files to compress.

Again this isn't really something to worry about, you would just compress all requests and responses.  There aren't any files.

> Proxy servers may interfere with gzip. Use the Vary HTTP response header to specify how proxies and caches should handle compressed content.

Vary is normally set by the HTTP server, not the client.  It is pretty advanced and I wouldn't avoid talking about it's use.

> Configure dynamic and static compression.

I don't understand what this means.

kelseiv:
Thank you, @danielnelson! Please see [7f88a3a](
https://github.com/influxdata/docs.influxdata.com/pull/2199/commits/7f88a3ab2b86f65aff0f00d84ffca27ee90c166c) and let me know if you have add'l edits.

PR #2225, opened 2019-04-18
change clone to https - more universal

karel-rehor:
Changed the clone url to https, which github recommends and is more universal - fewer proxy issues, no account required. 

See issue in sandbox

https://github.com/influxdata/sandbox/issues/63


sanderson:
Thanks @karel-rehor!

PR #2230, opened 2019-04-24
Modify alert match format

chrisbrowning:
If the entire `match` condition for checking tags is not wrapped in double quotes with escaped double quotes around the tag name, kapacitor complains due to an error converting YAML to JSON. I ran into this issue after consulting the docs, which do not properly reflect how this should be formatted. I confirmed that the format in this PR works through my own testing.

Others have also encountered this issue as well and found the same solution, per https://github.com/influxdata/kapacitor/issues/1621.

### DIFF
--- a/content/kapacitor/v1.5/working/alerts.md
+++ b/content/kapacitor/v1.5/working/alerts.md
@@ -159,7 +159,7 @@ match: level() >= WARNING
 Send events with the tag "host" equal to `s001.example.com` to the handler:
 
 ```yaml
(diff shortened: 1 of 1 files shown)


sanderson:
Thanks @chrisbrowning! Before we can merge community contributions, we just need you to sign the InfluxData [Contributor License Agreement (CLA)](https://www.influxdata.com/legal/cla/).

chrisbrowning:
👍  done!

sanderson:
Thanks! Merged.

PR #2257, opened 2019-05-27
feat: add new Erlang line encoder lib

Pouriya-Jahanbakhsh:


sanderson:
Thanks @Pouriya-Jahanbakhsh

PR #2305, opened 2019-06-27
Flux 0.33

sanderson:
This PR should not be released until InfluxDB 1.7.8 is released.


sanderson:
All these changes are included in the Flux 0.36.2 PR.

Issue #2302, opened 2019-06-25
Note the "index-version" config setting on TSI pages


gunnaraasen:
We do not mention the `index-version` config value to [enable TSI](https://docs.influxdata.com/influxdb/v1.7/administration/config/#index-version-inmem) on many of the different pages where we talk about TSI in the docs. We should also reference the [TSI conversion steps](https://github.com/influxdata/docs.influxdata.com/issues/2201) when referencing the config.

We do not make it obvious enough that this setting is required and not on by default. We should at least mention `index-version` on the TSI pages.

###### URL for relevant page?
https://docs.influxdata.com/influxdb/v1.7/concepts/tsi-details/#time-series-index-tsi-description
https://docs.influxdata.com/influxdb/v1.7/concepts/time-series-index/#time-series-index-tsi

###### What products and version are you using?
InfluxDB 1.7 and InfluxDB Enterprise 1.7

kelseiv:
thank you, @gunnaraasen!

PR #2306, opened 2019-06-28
add index-version 

kelseiv:
per #2302

TSI details
- add index-version = tsi1 
- add reference to TSI conversion steps/config for OSS and Enterprise 

TSI overview
- add reference to TSI details


kelseiv:
thank you, @gunnaraasen!

PR #2320, opened 2019-07-11
fix swapped descriptions for output/processor plugins

kevin-cantwell:
Addresses https://github.com/influxdata/docs.influxdata.com/issues/2321

I noticed that the descriptions for output and processor plugins seemed swapped in a bunch of places.

This change makes sure that:
1. Output plugins are always described as "write metrics to various destinations."
1. Processor plugins are always described as "transform, decorate, and filter metrics."


stevebang:
Closing -- sorry, @kevin-cantwell, for missing the PR. I noticed the issue and fixed it while I was in the docs.

PR #2350, opened 2019-08-01
Kapacitor query node

kelseiv:
Update docs per [Kapacitor-idiomatic way for running advanced-syntax InfluxQL functions](https://github.com/influxdata/kapacitor/issues/2222).

### DIFF
--- a/content/kapacitor/v1.5/nodes/query_node.md
+++ b/content/kapacitor/v1.5/nodes/query_node.md
@@ -13,7 +13,7 @@ menu:
 
 The `query` node defines a source and a schedule for processing batch data. Data is queried from InfluxDB, computed by the `query` node, and then passed into the data pipeline.
 
->**Note:** Kapacitor requires that non-negative values for aggregate functions (such as differences, derivatives, and so on) be computed outside of the `query` node.
+>**Note:** Kapacitor requires aggregate functions (such as differences, derivatives, and so on) be computed outside of the `query` node.
 
 Example:
 
@@ -27,13 +27,28 @@ batch
     .period(1m)
     .every(20s)
     .groupBy(time(10s), 'cpu')
-  | difference('max_usage')
-  | where(lambda: "difference" >= 0)
     ...
 ```
 
 In the example above, InfluxDB is queried every 20 seconds; the window of time returned
-spans 1 minute and is grouped into 10 second buckets. Non-negative values are computed before passing data to the `query` node.
+spans 1 minute and is grouped into 10 second buckets.
+
+The `query` node cannot contain grouping clauses or conditions on time. To compute an aggregate over groups, specify grouping and the aggregate itself using TICKScript. For example, the following script calculates the non-negative difference:
+
+```js
+batch
+  |query('''
+    SELECT mean("value")
+    FROM "telegraf"."default".cpu_usage_idle
+    WHERE "host" = 'serverA'
+  ''')
+    .period(1m)
+    .every(20s)
+    .groupBy(time(10s), 'cpu')
+  | difference('max_usage')
+  | where(lambda: "difference" >= 0)
+  ...
+```
 
 ### Constructor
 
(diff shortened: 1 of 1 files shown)


content/kapacitor/v1.5/nodes/query_node.md
```diff
@@ -11,11 +11,11 @@ menu:
     parent: nodes
 ---
 
-The `query` node defines a source and a schedule for processing batch data.
-The data is queried from an InfluxDB database and then passed into the data pipeline.
+The `query` node defines a source and a schedule for processing batch data. Data is queried from InfluxDB, computed by the `query` node, and then passed into the data pipeline.
 
-Example:
+>**Note:** Kapacitor requires that non-negative values for aggregate functions (such as differences, derivatives, and so on) be computed outside of the `query` node.
```
affo: `non-negative` is too restrictive. I would say:

```
Kapacitor requires aggregate functions (such as differences, derivatives, and so on) be computed outside of the `query` node
```

kelseiv:
@affo, please let me know if changes look OK, or if we need to tweak a bit. Thank you!

content/kapacitor/v1.5/nodes/query_node.md
```diff
@@ -27,12 +27,28 @@ batch
...
     .groupBy(time(10s), 'cpu')
-  ...
+    ...
 ```
 
-In the above example InfluxDB is queried every 20 seconds; the window of time returned
+In the example above, InfluxDB is queried every 20 seconds; the window of time returned
 spans 1 minute and is grouped into 10 second buckets.
 
+The `query` node cannot contain grouping clauses or conditions on time. To compute an aggregate over groups, specify grouping and the aggregate itself using TICKScript. For example, the following script calculates the non-negative difference:
```
affo: I would rewrite as:

If you want to use InfluxQL advanced syntax for functions in the `query` node, you will need to express part of the InfluxQL query in the TICKScript. Advanced syntax computes the nested aggregation over time buckets, and then applies the outer aggregate on the results.
For example, the following InfluxQL script calculates the non-negative difference of the 10-second mean of cpu usage for every cpu:

```
SELECT non_negative_difference(mean("value"))
  FROM "telegraf"."default"."cpu_usage_idle"
  WHERE "host" = 'serverA'
  GROUP BY time(10s), "cpu"
```

The `query` node cannot contain grouping clauses or conditions on time, so, the above is not valid InfluxQL for the `query` node. To achieve the result above, you must specify grouping and the outer aggregate using TICKScript:

```js
batch
  |query('''
    SELECT mean("value")
    FROM "telegraf"."default".cpu_usage_idle
    WHERE "host" = 'serverA'
  ''')
    .period(1m)
    .every(1m)
    .groupBy(time(10s), 'cpu')
  | difference('max_usage')
  | where(lambda: "difference" >= 0)
  ...
```

kelseiv: Thank you, @affo! Please see update and let me know if OK.

content/kapacitor/v1.5/nodes/query_node.md
```diff
@@ -27,12 +25,41 @@ batch
     .period(1m)
     .every(20s)
     .groupBy(time(10s), 'cpu')
+    ...
+```
+
+In the example above, InfluxDB is queried every 20 seconds; the window of time returned spans 1 minute and is grouped into 10 second buckets.
+
+>**Note:** Kapacitor requires aggregate functions (such as differences, derivatives, and so on) be computed outside of the `query` node. The `query` node cannot contain a grouping clause or condition on time.
```
affo: Could you remove this? It seems to be against what you said above.

kelseiv: Indeed, thank you @affo

kelseiv:
Removed note, thank you much @affo!

DarrenMendalka:
Great work on this one..   Looks good .

Issue #2338, opened 2019-07-22
Document the format of the http access_log


rbetts:
###### URL for relevant page?

https://docs.influxdata.com/influxdb/v1.7/administration/logs/

###### What products and version are you using?

InfluxDB 1.6+

###### Where did you look before opening the issue?

Above link.

Request to document the http access log format that we spew.  Example log line:

172.19.8.11,172.19.5.162 - - [21/Jul/2019:03:01:27 +0000] "GET /query?db=metrics&q=SELECT+MEAN%28value%29+as+average_cpu%2C+MAX%28value%29+as+peak_cpu+FROM+%22foo.load%22+WHERE+time+%3E%3D+now%28%29+-+1m+AND+org_id+%21%3D+%27%27+AND+group_id+%21%3D+%27%27+GROUP+BY+org_id%2Cgroup_id HTTP/1.0" 200 11450 "-" "Baz Service" d4ca9a10-ab63-11e9-8942-000000000000 9357049  

I _think_ the format is:

```
	return fmt.Sprintf(`%s - %s [%s] "%s %s %s" %s %s "%s" "%s" %s %d`,
		host,
		detect(username, "-"),
		start.Format("02/Jan/2006:15:04:05 -0700"),
		r.Method,
		uri,
		r.Proto,
		detect(strconv.Itoa(l.Status()), "-"),
		strconv.Itoa(l.Size()),
		detect(referer, "-"),
		detect(userAgent, "-"),
		r.Header.Get("Request-Id"),
		// response time, report in microseconds because this is consistent
		// with apache's %D parameter in mod_log_config
		int64(time.Since(start)/time.Microsecond))
```

From: `influxdb/services/httpd/response_logger.go`

timhallinflux:
Let's add this to the platform documentation section.

noramullen1:
@timhallinflux I'm wondering if it makes more sense for this to go in the influxdb logs documentation (https://docs.influxdata.com/influxdb/v1.7/administration/logs/) instead?

timhallinflux:
yeah...I think you are right.

PR #2359, opened 2019-08-06
Http access log format

noramullen1:
Addresses #2338


content/influxdb/v1.7/administration/logs.md
```diff
@@ -72,9 +72,21 @@ If using the package install on a `sysvinit` system, the config file for logrota
...
+Use the HTTP access log to log HTTP request traffic separately from the other InfluxDB log output.
+
+### HTTP access log format
+
+The following is an example of the HTTP access log format:
+
+```
+172.13.8.13,172.39.5.169 - - [21/Jul/2019:03:01:27 +0000] "GET /query?db=metrics&q=SELECT+MEAN%28value%29+as+average_cpu%2C+MAX%28value%29+as+peak_cpu+FROM+%22foo.load%22+WHERE+time+%3E%3D+now%28%29+-+1m+AND+org_id+%21%3D+%27%27+AND+group_id+%21%3D+%27%27+GROUP+BY+org_id%2Cgroup_id HTTP/1.0" 200 11450 "-" "Baz Service" d6ca5a13-at63-11o9-8942-000000000000 9337349
+```
+
```
rbetts: My goal in filing the issue was to document what the components of the log message mean - as otherwise it is just a bunch of random numbers :-)

noramullen1:
Thanks @rbetts. I've updated with a table. The one entry I wasn't sure about was referrer--will there ever be a referrer for this log?

content/influxdb/v1.7/administration/logs.md
```diff
@@ -72,9 +72,38 @@ If using the package install on a `sysvinit` system, the config file for logrota
...
+
+
+| Component                    | Example                                                                                                                      |
+|---                           |---                                                                                                                           |
+|Host                          |`172.13.8.13,172.39.5.169`                                                                                                    |
+|Time of log event             |`[21/Jul/2019:03:01:27 +0000]`                                                                                                |
+|Request method                |`GET`                                                                                                                         |
+|Username                      |`user`                                                                                                                        |
+|URI                           |`/query?db=metrics%26q=SELECT%20used_percent%20FROM%20%22telegraf.autogen.mem%22%20WHERE%20time%20%3E=%20now()%20-%201m%20	` |
```
timhallinflux: So, I'd like to see us expand on this field and how to break it down and make it more useful.

First, this is the HTTP API call being made.  For reference we should point to the existing HTTP API page which describes this in more detail.

Second, we should highlight that this field is URL encoded.  Tools like: https://www.url-encode-decode.com/ are handy to turn these strings into something a bit more usable.  For example, this string decodes to: /query?db=metrics&q=SELECT used_percent FROM "telegraf.autogen.mem" WHERE time >= now() - 1m 

this shows the database being accessed and the exact query being run.

content/influxdb/v1.7/administration/logs.md
```diff
@@ -89,7 +118,7 @@ Locate the `[http]` section of your InfluxDB configuration file and set the `acc
 
 ## Structured logging
 
-With InfluxDB 1.5, structured logging is supported and enable machine-readable and more developer-friendly log output formats. The two new structured log formats, `logfmt` and `json`, provide easier filtering and searching with external tools and simplifies integration of InfluxDB logs  with Splunk, Papertrail, Elasticsearch, and other third party tools.
+Structured logging enables machine-readable and more developer-friendly log output formats. The two structured log formats, `logfmt` and `json`, provide easier filtering and searching with external tools and simplifies integration of InfluxDB logs  with Splunk, Papertrail, Elasticsearch, and other third party tools.
```
timhallinflux: How about commenting on using InfluxDB too...as a potential target for these logs?

content/influxdb/v1.7/administration/logs.md
```diff
@@ -72,9 +72,38 @@ If using the package install on a `sysvinit` system, the config file for logrota
...
+```
+
+
+
+| Component                    | Example                                                                                                                      |
+|---                           |---                                                                                                                           |
+|Host                          |`172.13.8.13,172.39.5.169`                                                                                                    |
+|Time of log event             |`[21/Jul/2019:03:01:27 +0000]`                                                                                                |
+|Request method                |`GET`                                                                                                                         |
+|Username                      |`user`                                                                                                                        |
```
sanderson: Be sure to include `user` in the example line. It will go between `..."GET` and `/query?...`

content/influxdb/v1.7/administration/logs.md
```diff
@@ -72,9 +72,38 @@ If using the package install on a `sysvinit` system, the config file for logrota
...
+|Request protocol              |`HTTP/1.0` 	                                                                                                                  |
+|HTTP response code            |`200`                                                                                                                         |
+|Size of response in bytes     |`11450`                                                                                                                       |
+|Referrer                      |`-`                                                                                                                           |
+|User agent                    |`Baz Service`                                                                                                                 |
+|Request ID                    |`d4ca9a10-ab63-11e9-8942-000000000000`                                                                                        |
+|Response time in microseconds |9357049                                                                                                                       |
```
sanderson: Looking at this again, I don't think it's microseconds. This would be almost 2.5 hours in ms 😄.

andrew-influx:
Got a ping from Joe. Reviewed update. Looks good to me. The break down of the log example is good as well as the additions about the two formats. May be more to do but I think it's a good step above what we have. Added myself as a Reviewer.

noramullen1:
Thanks @andrew-influx ! Is there any other context that you think would be helpful for customers to have on the different pieces here? Or is it sufficient as-is?

andrew-influx:
Sufficient as is at this time.

Let me know if I can help additionally.

PR #2361, opened 2019-08-07
Improve ldap doc

kelseiv:
Please see questions


content/enterprise_influxdb/v1.7/administration/ldap.md
```diff
@@ -0,0 +1,143 @@
...
+
+To load your LDAP configuration file, run the following command:
+
+```bash
+influxd-ctl ldap set-config /path/to/ldap.toml
+```
+
+## Sample LDAP configuration
+
+The following is a sample configuration file that connects to a publicly available LDAP server.
```
kelseiv: I saw Mark's comment about 3 ways to connect to LDAP server...do we want add any of that detail here?
.....
There are 3 ways to connect to an LDAP server:
1. Unencrypted. We support this.
2. LDAP over SSL/TLS. We don't support this. That behavior of LDAP is considered deprecated.
3. Initial connection unencrypted, and then upgrade to TLS.  This is the current industry preference for encrypted LDAP connections. I think we support this - I know the library we use supports this - but I don't see any test coverage that proves we support it.

content/enterprise_influxdb/v1.7/administration/ldap.md
```diff
@@ -0,0 +1,143 @@
...
+
+A `DN` is the distinguished name that uniquely identifies an entry and describes its position in the directory information tree (DIT) hierarchy. The DN of an LDAP entry is similar to a file path on a file system. `DNs` refers to multiple DN entries.
+
+```toml
+# As long as you have an SSH tunnel to the AD server,
```
kelseiv: Active Directory server?
...many examples below are specific to active directory. Should we distinguish/provide examples of both LDAP and AD?

content/enterprise_influxdb/v1.7/administration/ldap.md
```diff
@@ -0,0 +1,143 @@
...
+  port = 3389
+
+  # Security mode for LDAP connection to this server.
+  # Defaults to "starttls", to use an initial unencrypted connection
+  # and upgrade to TLS as the first action against the server,
+  # per the LDAPv3 standard.
+  # Other options are "starttls+insecure" to behave the same as starttls
+  # but skip server certificate verification, or "none" to use an unencrypted connection.
+  # security = "starttls"
+  security = "none"
```
kelseiv: I saw Mark's comment about 3 ways to connect to LDAP server...do we want add any of that detail here?
.....
There are 3 ways to connect to an LDAP server:

Unencrypted. We support this.
LDAP over SSL/TLS. We don't support this. That behavior of LDAP is considered deprecated.
Initial connection unencrypted, and then upgrade to TLS. This is the current industry preference for encrypted LDAP connections. I think we support this - I know the library we use supports this - but I don't see any test coverage that proves we support it.

content/enterprise_influxdb/v1.7/administration/ldap.md
```diff
@@ -0,0 +1,143 @@
...
+  - Set the `[http]` configuration setting for `shared-secret`, or the corresponding environment variable `INFLUXDB_HTTP_SHARED_SECRET`. Default value is `""`.
+- (Optional) If you're enabling authentication on meta nodes, data nodes must also include the following configurations:
+
+  - `INFLUXDB_META_META_AUTH_ENABLED` environment variable, or `[http]` configuration setting `meta-auth-enabled`, is set to `true`. Default value is `false`. This value must be the same value as the meta node's `meta.auth-enabled` configuration.
+  - `INFLUXDB_META_META_INTERNAL_SHARED_SECRET`, or the corresponding `[meta]` configuration setting `meta-internal-shared-secret`, is set to `true`. Default value is `false`. This value must be the same value as the meta node's `meta.internal-shared-secret`.
+
+> **Note:** To use fine-grained authorization (FGA) with LDAP, you must map FGA roles to key-value pairs in the LDAP database. For more information, see [Fine-grained authorization in InfluxDB Enterprise](/enterprise_influxdb/v1.7/guides/fine-grained-authorization/).
```
kelseiv: I saw Mark's comment about LDAP.md in plutonium notes: “Role modification isn't very easy to do at the moment, as it requires directly accessing the meta HTTP API. Follow the FGA guide but don't bother adding users to the roles.” Do we want to remove 'adding users to roles' in the FGA docs?

content/enterprise_influxdb/v1.7/administration/ldap.md
```diff
@@ -0,0 +1,143 @@
...
+A `DN` is the distinguished name that uniquely identifies an entry and describes its position in the directory information tree (DIT) hierarchy. The DN of an LDAP entry is similar to a file path on a file system. `DNs` refers to multiple DN entries.
+
+```toml
+# As long as you have an SSH tunnel to the AD server,
+# this config should work out of the box.
+# Verify with `influxd-ctl ldap verify /path/to/this.toml`.
+enabled = true
+
+[[servers]]
+  # Assumes you have an SSH tunnel to the AD server.
```
kelseiv: same

kelseiv: OK to assume one has an SSH tunnel to the AD server?

content/enterprise_influxdb/v1.7/administration/ldap.md
```diff
@@ -0,0 +1,143 @@
...
+  # security = "starttls"
+  security = "none"
+
+  # Credentials to use when searching for a user or group.
+  bind-dn = "CN=readonly admin,OU=Users,OU=enterprisead,DC=enterprisead,DC=example,DC=com"
+  bind-password = "p@ssw0rd"
+
+  # Base DNs to use when applying the search-filter to discover an LDAP user.
+  search-base-dns = [
+    "OU=Users,OU=enterprisead,DC=enterprisead,DC=example,DC=com",
```
kelseiv: valid?

content/enterprise_influxdb/v1.7/administration/ldap.md
```diff
@@ -0,0 +1,143 @@
...
+  group-search-base-dns = [
+    "OU=Users,OU=enterprisead,DC=enterprisead,DC=example,DC=com",
+  ]
+
+  # LDAP filter to identify groups that a user belongs to.
+  # %s will be replaced with the user'\''s DN.
+  group-membership-search-filter = "(&(objectClass=group)(member=%s))"
+
+  # Attribute to use to determine the "group" in the group-mappings section.
+  group-attribute = "cn"
```
kelseiv: Preferred format: cn=read-only-admin or cn=readonly admin

content/enterprise_influxdb/v1.7/administration/ldap.md
```diff
@@ -0,0 +1,143 @@
...
+
+```
```
kelseiv: do we want to add sample LDIF file to docs?

content/enterprise_influxdb/v1.7/administration/ldap.md
```diff
@@ -7,119 +7,119 @@ menu:
     weight: 40
     parent: Administration
 ---
-InfluxDB Enterprise can be configured to query a Lightweight Directory Access Protocol(LDAP)-compatible directory service for determining user permissions and to synchronize this directory service into InfluxDB so that the remote directory service does not need to be queried for each request.
 
-## Requirements
+Configure InfluxDB Enterprise to use LDAP (Lightweight Directory Access Protocol) to:
+
+- Validate user permissions
+- Synchronize InfluxDB and LDAP so each LDAP request doesn't need to be queried
```
mark-rushakoff: Maybe: "... so each InfluxDB request doesn't query the LDAP server"?

content/enterprise_influxdb/v1.7/administration/ldap.md
```diff
@@ -7,119 +7,119 @@ menu:
...
 ```toml
-# As long as you have an SSH tunnel to the AD server,
-# this config should work out of the box.
-# Verify with `influxd-ctl ldap verify /path/to/this.toml`.
+
+  # As long as you have an SSH tunnel to the LDAP server,
+  # this configuration should work out of the box.
+  # Verify with `influxd-ctl ldap verify /path/to/this.toml`.
+
+  # Note, Windows Active Directory may not work with LDAP.
```
mark-rushakoff: We have tested with AD. I suppose I would prefer this to somehow call out that configuration is difficult and generally requires some expertise in communicating with LDAP, as even a subtle or small mistake can cause the configuration to break.

content/enterprise_influxdb/v1.7/administration/ldap.md
```diff
@@ -7,119 +7,119 @@ menu:
...
 The following is a sample configuration file that connects to a publicly available LDAP server.
 
 A `DN` is the distinguished name that uniquely identifies an entry and describes its position in the directory information tree (DIT) hierarchy. The DN of an LDAP entry is similar to a file path on a file system. `DNs` refers to multiple DN entries.
 
 ```toml
-# As long as you have an SSH tunnel to the AD server,
-# this config should work out of the box.
-# Verify with `influxd-ctl ldap verify /path/to/this.toml`.
+
+  # As long as you have an SSH tunnel to the LDAP server,
```
mark-rushakoff: I'm not sure where this comment came from (I very well may have written it > 1yr ago in some test config file, maybe).

Having an SSH tunnel to an LDAP server is probably unusual. Maybe it's fine for an AD server, or maybe I wrote that note before we implemented starttls security.

I think it would be fine to leave in the comment, clarifying that starttls is preferred, but if your LDAP server doesn't support that for some reason, using an SSH tunnel with `security=none` is an okay workaround.

content/enterprise_influxdb/v1.7/administration/ldap.md
```diff
@@ -7,119 +7,119 @@ menu:
...
-  # The test server is not set up with TLS. Trying to use starttls against it will not work.
-  # security = "starttls"
-  security = "none"
+  # Other options are "starttls+insecure", which behaves the same as starttls
+  # but skips the server certificate verification, and "none", which uses an unencrypted connection.
+  security = "starttls"
 
   # Credentials to use when searching for a user or group.
-  bind-dn = "CN=readonly admin,OU=Users,OU=enterprisead,DC=enterprisead,DC=example,DC=com"
+  bind-dn = "cn=readonly admin,ou=users,ou=enterprisead,dc=enterprisead,dc=example,dc=com"
```
mark-rushakoff: The space in `readonly admin` is fine; CN values in LDAP are allowed to contain spaces.

content/enterprise_influxdb/v1.7/administration/ldap.md
mark-rushakoff: Should be fine to drop the sample LDIF.

content/enterprise_influxdb/v1.7/administration/ldap.md
```diff
@@ -128,73 +128,39 @@ enabled = true
...
+
+  [[servers.group-mappings]]
+    group = "admin"
+    role = "administrator"
+
+  [[servers.group-mappings]]
+    group = "user"
+    role = "monitor"
+
+  # To do `admin` tasks, log in with the `administrator` role username and password.
```
mark-rushakoff: "Create the administrator role in influxdb, then log in as a member of the admin LDAP group" perhaps?

The influxdb role is a collection of permissions and a group of users who receive those permissions.

content/enterprise_influxdb/v1.7/administration/ldap.md
kelseiv: Thank you, @mark-rushakoff 
If AD works (regardless of complexity), I think we can remove this note. (I read that LDAP might not have worked being it was deprecated awhile ago). @andrew-influx thoughts?

Perhaps we specify the most common issues that break the configuration. For docs, we typically stay away from saying something may not work if should work other than user error.

content/enterprise_influxdb/v1.7/administration/ldap.md
kelseiv: @mark-rushakoff, I'll add: "...we recommend using the default security configuration, "starttls"."

If having an SSH tunnel is unusual **and** we can't specify why an LDAP server wouldn't support  "starttls", probably best to avoid detailing SSH. @andrew-influx thoughts?

Alternatively, if we must:
**Note:** If your LDAP server doesn't support the "starttls" configuration 
<for xyz reasons?> <how will they know?>, 
you may opt to use an SSH tunnel with the security configuration set to "none".

To set up an SSH tunnel to the LDAP/AD server, run:
`ssh -N -L 3389:ad_server:389 jumpbox_ip`

To add a tunnel-only user, run:
`useradd tunnel -m -d /home/tunnel -s /bin/true && mkdir -p /home/tunnel/.ssh && cat pubkey >> /home/tunnel/.ssh/authorized_keys`

content/enterprise_influxdb/v1.7/administration/ldap.md
```diff
@@ -7,119 +7,119 @@ menu:
...
-  * Set the `[http]` `auth-enabled` configuration setting, or corresponding environment variable `INFLUXDB_HTTP_AUTH_ENABLED`, to `true`. Default is `false`.
-* Configure the HTTP shared secret to validate requests using JSON web tokens (JWT) and sign each HTTP payload with the secret and username.
-  * Set the `[http]` configuration setting for `shared-secret`, or the corresponding environment variable `INFLUXDB_HTTP_SHARED_SECRET`. Default value is `""`.
+Update the following settings in each data node configuration file (/etc/influxdb/influxdb.conf):
 
-### Configure the InfluxDB Enterprise meta nodes
+- Provide an HTTP Basic Authentication header. See [Authentication and authorization in InfluxDB](/influxdb/v1.7/administration/authentication_and_authorization/) for details on using HTTP Basic Authentication with InfluxDB.
+- Provide a username and password as HTTP query parameters:
+  - `u`: username
+  - `p`: password
```
sanderson: These first two line items aren't part of the configuration file. They are different methods for authenticating with the API once authentication is enabled.

Using basic auth with `curl -u`:
```
curl -u username:password -XPOST "http://localhost:8086/..."
```

Using HTTP query parameters with curl:
```
curl -XPOST "http://localhost:8086/.../?u=username&p=password"
```

We recommend using basic auth. I see these are in the "Configure meta nodes" section as well. I think they can be moved to as section after configuring the data/meta nodes that just talks about using authentication credentials when interacting with InfluxDB.

content/enterprise_influxdb/v1.7/administration/ldap.md
```diff
@@ -7,119 +7,119 @@ menu:
...
-```
-
-## Sample LDAP configuration
+### Sample LDAP configuration
 
 The following is a sample configuration file that connects to a publicly available LDAP server.
 
 A `DN` is the distinguished name that uniquely identifies an entry and describes its position in the directory information tree (DIT) hierarchy. The DN of an LDAP entry is similar to a file path on a file system. `DNs` refers to multiple DN entries.
 
 ```toml
```
sanderson: I would wrap this code block in the `{{% truncate %}}` shortcode to make it collapsable. Not everyone will want to see the entire LDAP config file example.

~~~
{{% truncate %}}
```
--- code block ---
```
{{% /truncate %}}
~~~

content/enterprise_influxdb/v1.7/administration/ldap.md
kelseiv: @sanderson, it looks like line 32 is part the config file. Does it make sense to keep the reference to the "Authentication and authorization in InfluxDB" procedure?

...and then after the toml sample file, add a section for lines 33-35 called:

 "**Authenticate your InfluxDB connection <using LDAP?>**"

- To authenticate your connection, run the following command, replacing username and password as needed:

    `curl -u username:password -XPOST "http://localhost:8086/..."`

sanderson: The InfluxDB config file doesn't set the Basic Authentication header. You just set `auth-enabled` to `true`. This requires authentication (username and password) with all requests to InfluxDB. You can provide authentication credentials using a basic auth header or query parameters.

This approach isn't unique to LDAP. LDAP just changes how you manage users and permissions, not the actual authentication process.

content/enterprise_influxdb/v1.7/administration/ldap.md
kelseiv: Removed the SSH tunnel to LDAP server reference.

PR #2364, opened 2019-08-07
Add FAQ for file system restrictions

kelseiv:


kelseiv:
resolves influxdata/DAR#19

PR #2392, opened 2019-08-23
AWS marketplace docs

noramullen1:


sanderson:
I fixed the frontmatter so the nested docs now show up in the nav.

noramullen1:
@gunnaraasen -- I've decided to publish this for now so we have something out there in the meantime. Let me know of any urgent updates you have!

PR #2393, opened 2019-08-23
UTC support

### LINKED ISSUE #2391: Chronograf 1.7.13 -- UTC support
###### URL for relevant page?
Users can now toggle between the local time (as reported by their browser) and UTC (Coordinated Universal Time).  This is handy for people working in large organizations which span time zones.  For most folks working in these environments, they will use UTC to ensure that they are all looking at metrics/events using a coordinated time -- rather than trying to interpret each individual's local time.

This is available in three places:
Status/Landing page :
Screen Shot 2019-08-23 at 10 24 44 AM

zoomed in: Screen Shot 2019-08-22 at 4 20 45 PM

https://docs.influxdata.com/chronograf/v1.7/guides/querying-data
UTC support is available as part of 1.7.13  -- to access this from Data Explorer
Screen Shot 2019-08-22 at 11 54 38 AM

UTC support is also available on each dashboard page as well. 
Screen Shot 2019-08-22 at 11 55 00 AM

Perhaps that needs to be added here: https://docs.influxdata.com/chronograf/v1.7/guides/create-a-dashboard/#step-6-configure-your-dashboard

UTC should be indexed -- as this will be a topic folks will search on.

Chronograf 1.7.13

kelseiv:
Fixes #2391

### DIFF
--- a/content/chronograf/v1.7/guides/create-a-dashboard.md
+++ b/content/chronograf/v1.7/guides/create-a-dashboard.md
@@ -65,7 +65,7 @@ Chronograf supports many different [visualization types](/chronograf/latest/guid
 ### Step 5: Save your cell
 Click **Save** (the green checkmark icon) to save your cell.
 
-> ***Note:*** if you navigate away from this page without clicking Save, your work will not be saved.
+> ***Note:*** If you navigate away from this page without clicking Save, your work will not be saved.
 
 ### Step 6: Configure your dashboard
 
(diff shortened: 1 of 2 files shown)


content/chronograf/v1.7/guides/create-a-dashboard.md
```diff
@@ -79,23 +79,26 @@ Click **Save** (the green checkmark icon) to save your cell.
 * **Pan** over your cell data by pressing the shift key and clicking and dragging your mouse over the graph
 * **Reset** your cell by double-clicking your mouse in the cell window
 
-> Note:
-These tips only apply to the line, stacked, step-plot, and line+stat [visualization types](/chronograf/latest/guides/visualization-types/).
+    > **Note:**
+    These tips only apply to the line, stacked, step-plot, and line+stat [visualization types](/chronograf/latest/guides/visualization-types/).
```
sanderson: Does this render correctly as a note? I think the 2nd line may need a `>`, but I may be wrong.

kelseiv: yes, thank you for checking!

PR #2395, opened 2019-08-27
Reformat table

ducnt114:


sanderson:
Thanks for the contribution @ducnt114. Was there an issue with the formatting of the previous table that was causing to not appear correctly for you?

ducnt114:
@sanderson Table format was broken as you can see in https://docs.influxdata.com/influxdb/v1.7/concepts/storage_engine/

sanderson:
@ducnt114 It appears correctly for me. Can you take a screenshot of how it renders for you?

ducnt114:
This is how the table rendered on my macbook pro 13"
Screen Shot 2019-09-11 at 4 00 44 PM

PR #2400, opened 2019-08-28
add example to limit SHOW SERIES by time

kelseiv:


kelseiv:
resolves influxdata/DAR#37

PR #2415, opened 2019-09-17
Recommend Installing TICK Stack with Helm Stable Charts

rawkode:
Our [tick-charts](https://github.com/influxdata/tick-charts) repository is not maintained. It has been decided that we should adopt the upstream Helm Stable repositories and support them as much as we can.

CC @gunnaraasen


gunnaraasen:
Thanks @rawkode!

Issue #2366, opened 2019-08-09
unclear in documentation : Show tag keys with where time clause uses shard duration as finest granularity 


marcvandermeiren:
see this discussion on the influx community to see what I mean :-) 

https://community.influxdata.com/t/undocumented-behaviour-with-show-tag-keys-values/10805

kelseiv:
Thank you, @marcvandermeiren! 

Updated [docs](https://docs.influxdata.com/influxdb/v1.7/query_language/schema_exploration/#filter-meta-queries-by-time).

PR #2416, opened 2019-09-19
Tag keys shard duration

kelseiv:
- Addresses #2366 (unclear in documentation : Show tag keys with where time clause uses shard duration as finest granularity)
- Add how to filter schema data (meta queries) by time using the shard group duration.
- Add how to filter results for Grafana template variables.
- Minor edits.

### DIFF
--- a/content/influxdb/v1.7/query_language/schema_exploration.md
+++ b/content/influxdb/v1.7/query_language/schema_exploration.md
@@ -1142,7 +1142,7 @@ For more information, see the
 
 When you filter meta queries by time, you may see results outside of your specified time. Meta query results are filtered at the shard level, so results can be approximately as granular as your shard group duration. If your time filter spans multiple shards, you'll get results from all shards with points in the specified time range. To review your shards and timestamps on points in the shard, run `SHOW SHARDS`. To learn more about shards and their duration, see [recommended shard groups durations](/influxdb/v1.7/concepts/schema_and_data_layout/#shard-group-duration-recommendations).
 
-The example below shows how to filter `SHOW TAG KEYS` approximately by one hour using a 1h shard group duration. To filter other meta data, replace `SHOW TAG KEYS` with `SHOW TAG VALUES`, `SHOW SERIES`, `SHOW MEASUREMENTS`, `SHOW FIELD KEYS`, and so on.
+The example below shows how to filter `SHOW TAG KEYS` by approximately one hour using a 1h shard group duration. To filter other meta data, replace `SHOW TAG KEYS` with `SHOW TAG VALUES`, `SHOW SERIES`, `SHOW MEASUREMENTS`, `SHOW FIELD KEYS`, and so on.
 
 #### Example filtering `SHOW TAG KEYS` by time
 
--- a/content/influxdb/v1.7/query_language/schema_exploration.md
+++ b/content/influxdb/v1.7/query_language/schema_exploration.md
@@ -765,9 +765,6 @@ The `FROM`, `WHERE`, `LIMIT`, and `OFFSET` clauses are optional.
 The `WHERE` clause supports tag comparisons; field comparisons are not
 valid for the `SHOW TAG KEYS` query.
 
->** Note:** 
-To [find tag keys within a specified time interval](#find-tag-keys-within-shard-duration), you must specify a shard duration in the same time interval.
-
 Supported operators in the `WITH` and `WHERE` clauses:
 `=`&emsp;&nbsp;&thinsp;equal to
 `<>`&emsp;not equal to
--- a/content/influxdb/v1.7/query_language/schema_exploration.md
+++ b/content/influxdb/v1.7/query_language/schema_exploration.md
@@ -23,7 +23,7 @@ The following sections cover useful query syntax for exploring your [schema](/in
   </tr>
   <tr>
     <td><a href="#show-field-keys">SHOW FIELD KEYS</a></td>
-    <td><a href="#filter-schema-data-by-time">Filter schema data by time</a></td>
+    <td><a href="#filter-meta-queries-by-time">Filter meta queries by time</a></td>
     <td></td>
   </tr>
 </table>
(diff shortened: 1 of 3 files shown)


content/influxdb/v1.7/query_language/schema_exploration.md
```diff
@@ -1137,3 +1137,100 @@ Note that `SHOW FIELD KEYS` handles field type discrepancies differently from
...
+#### Example filtering `SHOW TAG KEYS` by time
+
+1. Specify a shard duration on a new database or [alter an existing shard duration](/influxdb/v1.7/query_language/database_management/#modify-retention-policies-with-alter-retention-policy). To specify a 1h shard duration when creating a new database, run the following command:
+```js
+> CREATE database mydb with duration 7d REPLICATION 1 SHARD DURATION 1h name myRP;
+```
+
+     > **Note:** The minimum shard duration is 1h.
+
+2. Verify the shard duration has the correct time interval (precision) by running the `show shards` command. The example below shows a shard duration with an hour precision.
```
sanderson: Based solely on "tradition", make `show shards` all caps. `SHOW SHARDS`.

content/influxdb/v1.7/query_language/schema_exploration.md
```diff
@@ -1137,3 +1137,100 @@ Note that `SHOW FIELD KEYS` handles field type discrepancies differently from
...
+```js
```
sanderson: This code block isn't javascript or flux. Either remove the syntax declaration or use `sh` for "shell."

content/influxdb/v1.7/query_language/schema_exploration.md
```diff
@@ -1137,3 +1137,100 @@ Note that `SHOW FIELD KEYS` handles field type discrepancies differently from
...
+> show shards
+name: mydb
+id database retention_policy shard_group start_time end_time expiry_time owners
+-- -------- ---------------- ----------- ---------- -------- ----------- ------
+> precision h
+```
```
sanderson: I'm guessing this code block isn't nested under the line item and may be breaking the numbering. You may need to indent it 4 spaces.

content/influxdb/v1.7/query_language/schema_exploration.md
```diff
@@ -1137,3 +1137,100 @@ Note that `SHOW FIELD KEYS` handles field type discrepancies differently from
...
+> show shards
+name: mydb
+id database retention_policy shard_group start_time end_time expiry_time owners
+-- -------- ---------------- ----------- ---------- -------- ----------- ------
+> precision h
+```  
+
+3. (Optional) Insert sample tag keys. This step is for demonstration purposes. If you already have tag keys (or other meta data) to search for, skip this step.
+
+    ```js
```
sanderson: Ether use `sh`, `sql`, or remove the syntax.

content/influxdb/v1.7/query_language/schema_exploration.md
```diff
@@ -1137,3 +1137,100 @@ Note that `SHOW FIELD KEYS` handles field type discrepancies differently from
...
+    ```
+
+4. To find tag keys within a shard duration, run one of the following commands:
+   
+    `SHOW TAG KEYS ON database-name <WHERE time clause>` OR
+     
+    `SELECT * FROM measurement <WHERE time clause>`
+  
+    The examples below use test data from step 3.
+    ```js
```
sanderson: Same as above.

content/influxdb/v1.7/query_language/schema_exploration.md
```diff
@@ -1137,3 +1137,102 @@ Note that `SHOW FIELD KEYS` handles field type discrepancies differently from
 `SELECT` statements.
 For more information, see the
 [How does InfluxDB handle field type discrepancies across shards?](/influxdb/v1.7/troubleshooting/frequently-asked-questions/#how-does-influxdb-handle-field-type-discrepancies-across-shards).
+
+### Filter schema data by time
```
timhallinflux: I think I would drop the word schema here and just use "meta queries"

content/influxdb/v1.7/query_language/schema_exploration.md
```diff
@@ -1137,3 +1137,102 @@ Note that `SHOW FIELD KEYS` handles field type discrepancies differently from
...
+
+To filter queries on schema (also known as meta queries) by a specified time period, say one hour, you must have a shard group duration of the same period (1h). This happens because meta queries return all data in a shard (not individual datapoints). If a shard includes a datapoint with a timestamp in a specified time period, the meta query returns results from that shard.
```
timhallinflux: I think the explanation here is accurate, but awkward.  There are two things going on here...one is the desired behavior and the other is the actual behavior.  So, you can certainly achieve the desired behavior if and only if the time range filter for your meta query matches up precisely with the shard time boundaries. The likelihood of that happening is extraordinarily low.  Therefore, it is probably best to explain this in the context of the actual behavior so that the user knows what to expect...and how to interpret the results.


Perhaps this should be put into some kind of "highlight box".  Essentially, the point here is that filtering meta queries by time results in a "fuzzy" match.  Meaning, that while filtering is possible, it is not strict in terms of the time boundary specified.  This is because the results are filtered at the shard level.  So, if your time filter spans multiple shards -- you will get all results from all shards for which the time range applies.  One of the questions that will arise from this is "how do I gain an understanding of the time ranges for the underlying shards" -- use show shards to see that.... provide a link to that doc.

content/influxdb/v1.7/tools/_index.md
```diff
@@ -52,5 +52,21 @@ Influx Inspect is a tool designed to view detailed information about on disk sha
 Grafana is a convenient dashboard tool for visualizing time series data.
 It was originally built for Graphite, modeled after Kibana, and since been updated to support InfluxDB.
 
-<dt> Because of the [changes](/influxdb/v0.11/concepts/010_vs_011/#breaking-api-changes) to the `SHOW SERIES` and `SHOW TAG VALUES` formats in InfluxDB 0.11, InfluxDB 1.3+ will not work with the Query Editor in Grafana 2.6.
-This issue does not affect existing queries and dashboards or users working with Grafana 3.0. </dt>
+> **Tip:** If you use Grafana template variables, you can limit results (for example, a list of hosts) by a specified period of time.
```
timhallinflux: This approach is not unique to Grafana.  Chronograf also has the same "template variable" concept.  We should document both..... the mechanism (of using a continuous query and landing the data into a new measurement) can be used  by both UI tools.  The "last mile" instructions for creating the template variables in Grafana and Chronograf are different.  In fact, I believe there are two ways that this can be done.... using the CQ approach and, you can create a variable which is of type `custom meta query` that allows you to specify the time range filter.  The standard meta query template does not allow for time range filtering.

PR #2457, opened 2019-10-25
Restructured platform monitoring docs/dashboards

sanderson:


samhld:
@sanderson a couple comments below:

1) Someone might have a different opinion than me but I think we should use `max()` as the agg function in the derivative functions we use:

        a) I think seeing peak periods is more important than average periods
        b) Nathaniel once told me that using `mean()` in a derivative doesn't make much sense.  I'd need him to explain that though--I just trust him.

2) In the "HTTP Requests Per Minute" cell, it looks like only `queryReq` is being selected.  We'd want to ask for `writeReq` as well if we want "HTTP Requests" overall.  Perhaps even `req` as well, to show a total...or use a stacked line graph (either way works).

sanderson:
Thanks @samhld. I changed most of the queries to use `max()`. The "HTTP Requests Per Minute" cell has multiple queries:

- Query requests
- Write requests
- Client errors
- Server errors

samhld:
@sanderson ohh, I missed that there were multiple queries :).  Carry on!

PR #2459, opened 2019-10-28
Influxdb 1.7.9

sanderson:


content/influxdb/v1.7/about_the_project/releasenotes-changelog.md
```diff
@@ -7,6 +7,20 @@ menu:
...
 ---
 
+## v1.7.9 [2019-10-27]
+
+### Bugfixes
+- Guard against compaction burst throughput limit.
+- Replace TSI compaction wait group with counter.
+- Update InfluxQL dependency.
+- Add option to authenticate debug/pprof and ping endpoints.
+- Honor insecure-skip-very even if custom tls config is specified.
```
timhallinflux: I think this should be verify? (instead of `very`)

Also, let's capitalize TLS...

content/influxdb/v1.7/administration/config.md
```diff
@@ -654,11 +654,29 @@ Useful for troubleshooting and monitoring.
 
 Environment variable: `INFLUXDB_HTTP_PPROF_ENABLED`
 
+#### `pprof-auth-enabled = false`
```
timhallinflux: I find this a bit difficult to reconcile with the HTTP API guide: https://docs.influxdata.com/influxdb/v1.7/tools/api/

I believe that this setting protects the `/debug` portion of the API -- which will require users who enable this to supply creds for `/debug/pprof`, `/debug/requests` and `/debug/vars`.  The last one is the primary way that users extract telemetry data via Telegraf and the InfluxDB v1 input plugin.  I added: https://github.com/influxdata/telegraf/issues/6588

But, I think we need to describe this behavior further.  

Also, I'm not clear about the section below...and the /net/http/pprof binding to localhost:6060.  Seems like we should confirm this behavior.  The access is typically through the HTTP API which is exposed on :8086

sanderson: I updated the section about pprof auth.

Here's the source code for the call to `localhost:6060`: https://github.com/influxdata/influxdb/blob/1.7/cmd/influxd/run/command.go#L97-L102

PR #2462, opened 2019-10-29
Update writing_data.md

dragoshenron:
Fixed header for gzip compression


sanderson:
Thank your for your contribution @dragoshenron, but in the case of `gzip`, `application/` is not required. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding

PR #2481, opened 2019-11-05
Add docs for Kapacitor now() function

mpicker0:
I recently came upon Kapacitor's `now()` function via [this discussion](https://community.influxdata.com/t/solved-compare-a-date-field-with-now-in-kapacitor/1746), and would like to see it reflected in the docs.

This is my first PR against this repo; I tried to sign the [CLA](https://www.influxdata.com/legal/cla/) but the bottom half of the page is blank ~~and there seems to be no way to sign it~~.  Never mind, I think I got it.


mpicker0:
In that table, all the functions take a _time_ parameter and refer to the `time` field in the data.  The `now()` function seems like it should fall outside that; it has no parameters and instead returns a _time_ datatype.

Issue #2498, opened 2019-11-07
Add warning about writing fields and tags with same name


sanderson:
We need to call out that adding fields and tags with the same name will result in duplicate tags and the inability to select the tag.

Example from the community:
https://community.influxdata.com/t/why-is-my-tag-get-duplicated-suddenly-and-i-cant-select-using-the-tag/11678

We're waiting on possible solutions from @influxdata/support, but right now, it appears the only solution is to drop the affected shards.

###### URL for relevant page?
I'm not sure where best to land this. Maybe in the line protocol and client library docs. We may need to do an audit of Telegraf plugins to identify which ones this problem could occur with (input plugins where users structure the data like `csv` or `file`).

gunnaraasen:
Check out the [`::tag` and `::field` syntax](https://docs.influxdata.com/influxdb/v1.7/query_language/data_exploration/#select-specific-tags-and-fields-from-a-single-measurement-and-provide-their-identifier-type) in InfluxQL.

Also, there's [this troubleshooting section](https://docs.influxdata.com/influxdb/v1.7/troubleshooting/frequently-asked-questions/#identifier-names) about tags and fields with the same name.

gunnaraasen:
I agree we should make warnings about this more visible and put a note on the line protocol page.

kelseiv:
Thank you all! Content's merged.

PR #2510, opened 2019-11-14
Add info about duplicate keys

kelseiv:
- To address [duplicate field/tag keys issue](https://github.com/influxdata/docs.influxdata.com/issues/2498), add content to "Tag and field key with the same name" (line 619). 
- Other updates in this PR:

   -  Rewrote intro for "Why do my queries return no data or partial data?"
   -  Updated subheadings
   - Renamed "Identifier names" to "Tag and field key with the same name"
   - Reworded info about querying the wrong RP

### DIFF
--- a/content/influxdb/v1.7/troubleshooting/frequently-asked-questions.md
+++ b/content/influxdb/v1.7/troubleshooting/frequently-asked-questions.md
@@ -618,9 +618,62 @@ If your `SELECT` query includes a [`GROUP BY time()` clause](/influxdb/v1.7/quer
 
 ### Tag and field key with the same name
 
-Avoid using the same name for a tag and field key. If you inadvertently add the same name for a tag and field key, and then query both keys together, the query results show the second key (tag or field) appended with `_1` (also visible as the column header in Chronograf).
+Avoid using the same name for a tag and field key. If you inadvertently add the same name for a tag and field key, and then query both keys together, the query results show the second key queried (tag or field) appended with `_1` (also visible as the column header in Chronograf). To query a tag or field key appended with `_1`, drop the appended `_1` and include the syntax `::tag` or `::field`.
 
-Additionally, queries to the tag key don't return tag values. For example, if you have a tag key named `cpu` and write a field key named `cpu` to the same measurement, the tag key is renamed `cpu_1` and cannot be queried.
+#### Example
+
+Write the following points to create both a field and tag key with the same name `leaves`:
+
+```bash
+# create the `leaves` tag key
+`insert grape,leaves=species leaves=6`
+
+#create the `leaves` field key
+`insert grape leaves=5`
+```
+
+If you view both keys, you'll notice that neither key includes `_1`:
+
+```bash
+    > SHOW TAG KEYS
+    name: grape
+    tagKey
+    ------
+    leaves
+
+    > SHOW FIELD KEYS
+
+    name: grape
+    fieldKey   fieldType
+    ------     ---------
+    leaves     float
+```
+
+However, when you query the `grape` measurement, you'll see the `leaves` tag key has an appended `_1`:
+
+```bash
+   SELECT * FROM <database_name>.<retention_policy>."grape"
+
+   name: grape
+   time                leaves     leaves_1
+   ----                --------   ----------
+   1574128162128468000 first      species
+   1574128238044155000
+```
+
+To query a duplicate key name, you **must drop** `_1` **and include** `::tag` or `::field` after the key:
+
+```bash
+   SELECT "leaves"::tag, "leaves"::field FROM <database_name>.<retention_policy>."grape"
+
+   name: grape
+   time                leaves     leaves_1
+   ----                --------   ----------
+   1574128162128468000 species    6.00
+   1574128238044155000            5.00
+```
+
+Therefore, queries that reference `leaves_1` don't return values.
 
 > **Warning:** If you inadvertently add a duplicate key name, follow the steps below in "Remove a duplicate key." Because of memory requirements, if you have large amounts of data, we recommend chunking your data (while selecting it) by a specified interval (for example, date range) to fit the allotted memory.
 
--- a/content/influxdb/v1.7/troubleshooting/frequently-asked-questions.md
+++ b/content/influxdb/v1.7/troubleshooting/frequently-asked-questions.md
@@ -602,7 +602,7 @@ The most common reasons why your query returns no data or partial data:
 - [SELECT query includes `GROUP BY time()`](#select-query-includes-group-by-time) (partial data before `now()` returned)
 - [Tag and field key with the same name](#tag-and-field-key-with-the-same-name)
 
-### Querying wrong retention policies
+### Querying the wrong retention policy
 
 InfluxDB automatically queries data in a database’s `DEFAULT` retention policy](/influxdb/v1.7/concepts/glossary/#retention-policy-rp) (RP). If your data is stored in another RP, you must specify the RP in your query to get results.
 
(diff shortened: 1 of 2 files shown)


content/influxdb/v1.7/troubleshooting/frequently-asked-questions.md
```diff
@@ -595,35 +595,48 @@ time                    sunflowers                 time                  mean
...
+- [SELECT query includes `GROUP BY time()`](#select-query-includes-group-by-time) (partial data before `now()` returned)
+- [Tag and field key with the same name](#tag-and-field-key-with-the-same-name)
 
-The first and most common explanation involves [retention policies](/influxdb/v1.7/concepts/glossary/#retention-policy-rp) (RP).
-InfluxDB automatically queries data in a database’s `DEFAULT` RP.
-If your data is stored in an RP other than the `DEFAULT` RP, InfluxDB won’t return any results unless you specify the alternative RP.
+### Querying wrong retention policies
 
-### Tag keys in the SELECT clause
+InfluxDB automatically queries data in a database’s `DEFAULT` retention policies](/influxdb/v1.7/concepts/glossary/#retention-policy-rp) (RP). If your data is stored in another RP, you must specify the RP in your query to get results.
```
gunnaraasen: >  database’s `DEFAULT` retention policies

Remove the s on policies. Only one can be default.

content/influxdb/v1.7/troubleshooting/frequently-asked-questions.md
```diff
@@ -595,35 +595,48 @@ time                    sunflowers                 time                  mean
...
 
-The final common explanation involves [schemas](/influxdb/v1.7/concepts/glossary/#schema) with [fields](/influxdb/v1.7/concepts/glossary/#field) and [tags](/influxdb/v1.7/concepts/glossary/#tag) that have the same key.
-If a field and tag have the same key, the field will take precedence in all queries.
-You’ll need to use the [`::tag` syntax](/influxdb/v1.7/query_language/data_exploration/#description-of-syntax) to specify the tag key in queries.
+(By default, most [`SELECT` queries](/influxdb/v1.7/query_language/data_exploration/#the-basic-select-statement) query data with timestamps between `1677-09-21 00:12:43.145224194` and `2262-04-11T23:47:16.854775806Z` UTC.)
+
+### Tag and field key with the same name
+
+Avoid using the same name for a tag and field key. Adding a duplicate key name will rename the original key by appending the key name with `_1`. Queries to the renamed key don't return results. For example, if you have a tag key named `cpu` and write a field key named `cpu` to the same measurement, the tag key is renamed `cpu_1` and cannot be queried.
```
gunnaraasen: > the tag key is renamed `cpu_1` and cannot be queried

I'm not sure "renamed" is the right word here. Tag keys are written with the original tag key on the write path even if there's a duplicate field key. The `_x` suffix is added on the query-side when the tag and field are queried together.

content/influxdb/v1.7/write_protocols/line_protocol_reference.md
```diff
@@ -274,18 +263,23 @@ In those cases, `time` does not require double quotes in queries.
...
+A point is uniquely identified by the measurement name, tag set, field set, and timestamp
+
+If you write a point to a series with a timestamp that matches an existing point, the field set becomes a union of the old and new field set, and conflicts favor the new field set.
 
 For a complete example of this behavior and how to avoid it, see
 [How does InfluxDB handle duplicate points?](/influxdb/v1.7/troubleshooting/frequently-asked-questions/#how-does-influxdb-handle-duplicate-points)
 
+### Duplicate keys
+
+If you have a tag key called `location` and write a field key called `location` to the same measurement, the `location` tag key is **renamed** `location_1`. To retrieve data from the renamed tag key `location_1`, you **must** use the InfluxQL `::tag` syntax in your query, for example:
```
gunnaraasen: See comment about "rename" above. I think this implies to reader that values are save differently when it's really just a limitation of InfluxQL's expressiveness.

content/influxdb/v1.7/write_protocols/line_protocol_reference.md
```diff
@@ -274,18 +263,23 @@ In those cases, `time` does not require double quotes in queries.
...
+
+```sql
+  SELECT "<location_1>"::tag
```
gunnaraasen: Might as well make this a valid query.

content/influxdb/v1.7/write_protocols/line_protocol_reference.md
Anaisdg: Maybe two line protocol examples could be used here to be more illustrative? I don't think the tag key is "renamed" from the entire bucket/database.  It's just changed from the time you started writing the wrong points. I don't know if that's clear or even important, but I personally like examples: 
Previously inserted point in line protocol:
mymeasurement, location = california field = 1 timestamp
Previous Tag Keys:
location
New point:
mymeasurement,  location = california  location = california timestamp
New Tag keys
location 
location_1

kelseiv: @Anaisdg, looks like the field/key name never changes (even after "one writes the wrong points"). After one writes a duplicate key, queries return the appended `_1` (may be a tag or field key, whichever is queried second) and the tag key value cannot be queried. For example, you'll see the appended `_1` if you run `SELECT * FROM` but if you run `SHOW TAG KEYS ON`, you'll see the tag keys w/o the appended `_1`.  I'll include examples and expand the steps!

content/influxdb/v1.7/troubleshooting/frequently-asked-questions.md
```diff
@@ -595,35 +595,114 @@ time                    sunflowers                 time                  mean
...
+
+Write the following points to create both a field and tag key with the same name `leaves`:
+
+```bash
+# create the `leaves` tag key
+`insert grape,leaves=species leaves=6`
+
+#create the `leaves` field key
+`insert grape leaves=5`
+```
```
sanderson: - It's not clear where these should be run. This is how you write data using InfluxQL, so users can either do this in the Data Explorer or using the `influx` shell.
- Change the syntax highlighting from `bash` to `sql` (unless you ask them to run these from the `influx` shell).
- Remove the ticks inside the codeblock (lines 629 and 632).
- Make `insert` all-caps (`INSERT`). It's not necessary but matches other InfluxQL command examples.

content/influxdb/v1.7/troubleshooting/frequently-asked-questions.md
```diff
@@ -595,35 +595,114 @@ time                    sunflowers                 time                  mean
...
+    ------
+    leaves
+
+    > SHOW FIELD KEYS
+
+    name: grape
+    fieldKey   fieldType
+    ------     ---------
+    leaves     float
+```
```
sanderson: - This example would indicate that you're running the commands from the `influx` shell. I'd suggest just telling uses to use it that the begging of the example.
- Remove the indentation in the codeblock. These don't need to be indented.

content/influxdb/v1.7/troubleshooting/frequently-asked-questions.md
```diff
@@ -595,35 +595,114 @@ time                    sunflowers                 time                  mean
...
+However, when you query the `grape` measurement, you'll see the `leaves` tag key has an appended `_1`:
+
+```bash
+   SELECT * FROM <database_name>.<retention_policy>."grape"
+
+   name: grape
+   time                leaves     leaves_1
+   ----                --------   ----------
+   1574128162128468000 first      species
+   1574128238044155000
```
sanderson: Remove the indentation.

content/influxdb/v1.7/troubleshooting/frequently-asked-questions.md
```diff
@@ -595,35 +595,114 @@ time                    sunflowers                 time                  mean
...
+To query a duplicate key name, you **must drop** `_1` **and include** `::tag` or `::field` after the key:
+
+```bash
+   SELECT "leaves"::tag, "leaves"::field FROM <database_name>.<retention_policy>."grape"
+
+   name: grape
+   time                leaves     leaves_1
+   ----                --------   ----------
+   1574128162128468000 species    6.00
+   1574128238044155000            5.00
```
sanderson: Remove the indentation.

content/influxdb/v1.7/troubleshooting/frequently-asked-questions.md
```diff
@@ -595,35 +595,114 @@ time                    sunflowers                 time                  mean
...
+```
+
+Therefore, queries that reference `leaves_1` don't return values.
+
+> **Warning:** If you inadvertently add a duplicate key name, follow the steps below in "Remove a duplicate key." Because of memory requirements, if you have large amounts of data, we recommend chunking your data (while selecting it) by a specified interval (for example, date range) to fit the allotted memory.
```
sanderson: - Since this is a warning, use the `{{% warn %}}` shortcode  or `<dt>...</dt>` element to format this in an orange block.
- Link "Remove a duplicate key" to the actual heading.

content/influxdb/v1.7/troubleshooting/frequently-asked-questions.md
```diff
@@ -595,35 +595,114 @@ time                    sunflowers                 time                  mean
...
+# move data from temporary measurement back to original measurement you just dropped
+SELECT * INTO "original_measurement" FROM "temporary_measurement" GROUP BY *
+
+# verify the field keys and tags keys were successfully moved back to the original measurement
+SELECT * FROM "original_measurement"
+
+# drop temporary measurement
+DROP MEASUREMENT "temporary_measurement"
+
+```
```
sanderson: - Use `sql` syntax highlighting instead of `bash`.
- Update the comment characters (`#`) to use [InfluxQL comment characters](https://docs.influxdata.com/influxdb/v1.7/query_language/spec/#comments)
- Wrap the comments so they're no more than 80 characters wide.

content/influxdb/v1.7/troubleshooting/frequently-asked-questions.md
kelseiv: thank you @gunnaraasen, took out the "renamed" part. And clarified they need to drop `_1` when running a query

content/influxdb/v1.7/troubleshooting/frequently-asked-questions.md
kelseiv: added reference to launch `influx` 
removed ticks and capitalized insert

content/influxdb/v1.7/troubleshooting/frequently-asked-questions.md
kelseiv: - used sql, but also referred to launch `influx` again...we should prob touch base on this tomorrow so we all know what our standard is going forware- there's some inconsistency throughout the doc in general (both style and direction for where to write data)
- used the mulit-line influxql comments for consistency

kelseiv:
Addresses /influxdata/DAR/issues/11, specifically the section "Tag and field key with the same name" that starts on line 619

PR #2544, opened 2019-12-11
Document signature verification for Telegraf binaries

pierwill:


pierwill:
(Will probably revise as an ordered list—)

content/telegraf/v1.12/introduction/installation.md
```diff
@@ -207,6 +207,35 @@ sudo systemctl start telegraf
 {{< /tab-content-container >}}
 {{< /tab-labels >}}
 
+### Verify the authenticity of downloaded binary
```
sanderson: I think we can call out that this is an optional step, but recommended.

content/telegraf/v1.12/introduction/installation.md
```diff
@@ -207,6 +207,35 @@ sudo systemctl start telegraf
...
+
+InfluxData cryptographically signs each Telegraf binary release.
+To verify the signature of your download, first download and import our public key:
```
sanderson: We need to call out that users will need to install `gpg` and point to installation instructions for it.

sanderson:
Whatever changes you make in `content/telegraf/v1.12/` need to be ported to `content/telegraf/v1.13` on the `telegraf-1.13` branch. There is currently a PR open for this branch.

Issue #2575, opened 2020-01-10
Update Kubernetes-related documentation


pierwill:
The content on this page could use updating and editing: https://docs.influxdata.com/platform/integrations/kubernetes/#should-i-run-influxdb-in-kubernetes

We should direct users interested in exploring InfluxDB + Kubernetes to the v2 docs.

PR #2576, opened 2020-01-10
Link to v2 Kubernetes instructions from platform docs

pierwill:
Work toward #2575

### DIFF
--- a/content/platform/integrations/kubernetes.md
+++ b/content/platform/integrations/kubernetes.md
@@ -12,8 +12,13 @@ has become a popular way to deploy and manage containers across multiple servers
 
 ![InfluxDB Kubernetes Logos](/img/platform/flux-kube.png)
 
-There are several ways use the InfluxData Platform (also known as the TICK
-stack) with Kubernetes:
+{{% note %}}
+This page is about using Kubernetes with the TICK stack and 1.x versions of InfluxDB.
+
+To start experimenting with InfluxDB 2.0 and Kubernetes, go to the **Kubernetes** section of the [Get Started page for InfluxDB 2.0](https://v2.docs.influxdata.com/v2.0/get-started/).
+{{% /note %}}
+
+There are several ways use the InfluxData Platform with Kubernetes:
 
 - [Monitor Kubernetes](#monitor-kubernetes)
   - [kube-influxdb Kubernetes monitoring project](#kube-influxdb-kubernetes-monitoring-project)


content/platform/integrations/kubernetes.md
```diff
@@ -27,6 +27,10 @@ stack) with Kubernetes:
   - [How is the InfluxData Platform (TICK) different from Prometheus?](#how-is-the-influxdata-platform-tick-different-from-prometheus)
   - [Should I run InfluxDB in Kubernetes?](#should-i-run-influxdb-in-kubernetes)
 
+## Quick start with minikube
+
+To start experimenting right away with the latest InfluxDB and Kubernetes, visit [Get started with InfluxDB](https://v2.docs.influxdata.com/v2.0/get-started/) and click **Kubernetes**.
+
```
sanderson: We need to make sure it's clear that this method is only for InfluxDB v2. Maybe add it into the heading, link text, etc. 

This would be very confusing if somebody is in the v1.x docs looking for ways to get started with InfluxDB and Kubernetes and this link takes them to 2.0 without indication that it's only meant for 2.0. Many 1.x users are not ready to make the jump to 2.0, even with k8s.

gunnaraasen: Agreed, we need to make it clear what is v1 or v2. I think it would probably be best to make this a blurb at the top informing users that there's a different set of k8s resources for v2 and this content only applies to v1.

pierwill:
Pushed another version. How does this look?

content/platform/integrations/kubernetes.md
```diff
@@ -12,8 +12,13 @@ has become a popular way to deploy and manage containers across multiple servers
 
 ![InfluxDB Kubernetes Logos](/img/platform/flux-kube.png)
 
-There are several ways use the InfluxData Platform (also known as the TICK
-stack) with Kubernetes:
+{{% note %}}
+This page is about using Kubernetes with the TICK stack and 1.x versions of InfluxDB.
+
+To start experimenting with InfluxDB version 2 and Kubernetes, visit [Get started with InfluxDB](https://v2.docs.influxdata.com/v2.0/get-started/) and click **Kubernetes**.
```
gunnaraasen: "click Kubernetes" reads weird to me. How about this wording?

> ...go to the **Kubernetes** section of the [Get Started page for InfluxDB v2](https://v2.docs.influxdata.com/v2.0/get-started/).

sanderson: ~InfluxDB version 2~ -> InfluxDB 2.0

PR #2580, opened 2020-01-13
update per DAR 65

kelseiv:
Updated per DAR 65 to recommend folks use align() when ingesting data in irregular intervals.

### DIFF
--- a/content/kapacitor/v1.5/nodes/window_node.md
+++ b/content/kapacitor/v1.5/nodes/window_node.md
@@ -133,6 +133,7 @@ from 12:05 to 12:10).
 ```js
 window.align()
 ```
+> Note: When ingesting data at irregular intervals, we recommend using `window.align()` to prevent duplicate values with the same timestamp or unexpected additional values.
 
 <a class="top" href="javascript:document.getElementsByClassName('article-heading')[0].scrollIntoView();" title="top"><span class="icon arrow-up"></span></a>
 
--- a/content/kapacitor/v1.5/nodes/window_node.md
+++ b/content/kapacitor/v1.5/nodes/window_node.md
@@ -128,7 +128,7 @@ If the `align` property isn't used to modify the `window` node, the window align
 ```js
 window.align()
 ```
-> Note: When ingesting data at irregular intervals, we recommend using `window.align()` to prevent duplicate values with the same timestamp or unexpected additional values.
+> Note: When ingesting data at irregular intervals, we recommend using `window.align()` to group data.
 
 <a class="top" href="javascript:document.getElementsByClassName('article-heading')[0].scrollIntoView();" title="top"><span class="icon arrow-up"></span></a>
 
--- a/content/kapacitor/v1.5/nodes/window_node.md
+++ b/content/kapacitor/v1.5/nodes/window_node.md
@@ -119,16 +119,11 @@ Property methods modify state on the calling node.
 They do not add another node to the pipeline, and always return a reference to the calling node.
 Property methods are marked using the `.` operator.
 
-
 ### Align
 
-If the `align` property is not used to modify the `window` node, then the
-window alignment is assumed to start at the time of the first data point it receives.
-If `align` property is set, the window time edges
-will be truncated to the `every` property (For example, if a data point's time
-is 12:06 and the `every` property is `5m` then the data point's window will range
-from 12:05 to 12:10).
+Set the `align` property to truncate the window time edges to the `every` property. For example, if a data point's time is 12:06 and the `every` property is `5m` then the data point's window ranges from 12:05 to 12:10).
 
+If the `align` property isn't used to modify the `window` node, the window alignment starts at the time the first data point is received.
 
 ```js
 window.align()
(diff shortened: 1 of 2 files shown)


content/kapacitor/v1.5/nodes/window_node.md
```diff
@@ -119,20 +119,16 @@ Property methods modify state on the calling node.
...
 
-If the `align` property is not used to modify the `window` node, then the
-window alignment is assumed to start at the time of the first data point it receives.
-If `align` property is set, the window time edges
-will be truncated to the `every` property (For example, if a data point's time
-is 12:06 and the `every` property is `5m` then the data point's window will range
-from 12:05 to 12:10).
+Set the `align` property to truncate the window time edges to the `every` property. For example, if a data point's time is 12:06 and the `every` property is `5m` then the data point's window ranges from 12:05 to 12:10).
 
+If the `align` property isn't used to modify the `window` node, the window alignment starts at the time the first data point is received.
```
kelseiv: Does `align()` take both start and stop times? If not, what determines the stop time?

content/kapacitor/v1.5/nodes/window_node.md
```diff
@@ -119,20 +119,16 @@ Property methods modify state on the calling node.
...
-
 ### Align
 
-If the `align` property is not used to modify the `window` node, then the
-window alignment is assumed to start at the time of the first data point it receives.
-If `align` property is set, the window time edges
-will be truncated to the `every` property (For example, if a data point's time
-is 12:06 and the `every` property is `5m` then the data point's window will range
-from 12:05 to 12:10).
+Set the `align` property to truncate the window time edges to the `every` property. For example, if a data point's time is 12:06 and the `every` property is `5m` then the data point's window ranges from 12:05 to 12:10).
```
kelseiv: Does `align()` always start at (first?) data point -1m...if not, how does align determine the start time?

nathanielc: The start time is truncated to the every time, so if the every time is `1h` then the data point's time is truncated to the hour, so `12:59` would mean the start time is `12:00`.

content/kapacitor/v1.5/nodes/window_node.md
nathanielc: `stop = start + period` So if the start time is truncated then the stop time is likely also truncated because the `period` duration is almost always a multiple of the `every` duration. That said `align` doesn't directly set the `stop` time rather the equation `stop = start + period ` is always used.

content/kapacitor/v1.5/nodes/window_node.md
```diff
@@ -119,20 +119,16 @@ Property methods modify state on the calling node.
...
 
+If the `align` property isn't used to modify the `window` node, the window alignment starts at the time the first data point is received.
 
 ```js
 window.align()
 ```
+> Note: When ingesting data at irregular intervals, we recommend using `window.align()` to prevent duplicate values with the same timestamp or unexpected additional values.
```
nathanielc: Can you clarify this note? Window will not create duplicate values even if `align` is not used.

content/kapacitor/v1.5/nodes/window_node.md
kelseiv: Thank you @Nathaniel! Will a data point falling on the interval switch be included with prior or next window interval? 

For example, if the data point is 12:06 and the every time is:
3m, the window interval is truncated to 12:03-12:06 or 12:06-12:09?
1h, the window interval is truncated from 12-12:59 or 12-1?

content/kapacitor/v1.5/nodes/window_node.md
kelseiv: @nathanielc this note originated from a [DAR request](https://github.com/influxdata/DAR/issues/65)

the customer said "We see duplicate values..." that were resolved by using `align`...they also referenced a community issue: https://community.influxdata.com/t/stream-tickscript-why-do-i-see-duplicate-values/11649

Perhaps there's a more accurate way to describe this issue? Thank you!

content/kapacitor/v1.5/nodes/window_node.md
nathanielc: The bounds are inclusive start and exclusive stop. 

>  For example, if the data point is 12:06 and the every time is:
3m, the window interval is truncated to 12:03-12:06 or 12:06-12:09?
1h, the window interval is truncated from 12-12:59 or 12-1?

So 12:06 lands in 12:06-12:09

For 1h the interval is from 12:00-1:00, so the time 12:59:59.9999 is in the interval, but 1:00 is outside the interval.

content/kapacitor/v1.5/nodes/window_node.md
nathanielc: Yeah that community question is a bit misleading, the `httpOut` node will only show one point for stream data even if there are duplicate points. Its a cache of the latest point, so any duplicate data in the stream would be hidden. To correctly debug the issues the `log` node should be used. Its likely the user has duplicate data to start with but httpOut hides that fact. Once the data has been windowed then the httpOut node caches the most recent window, that is why it appears to be creating duplicate data. But if a log node is used before the window you would see that the duplicate data exists even before the window. 

Windows do not duplicate data with or without align.

PR #2582, opened 2020-01-13
Short-term warning in OSS/Enterprise migration guide

sanderson:
This is a short-term warning to prevent users from using the current OSS to Enterprise migration guide until InfluxDB 1.7.10 is available. Once it becomes available, we'll remove this warning and update the guide to reflect the new upgrade process.

### DIFF
--- a/content/enterprise_influxdb/v1.5/guides/migration.md
+++ b/content/enterprise_influxdb/v1.5/guides/migration.md
@@ -9,6 +9,19 @@ menu:
     parent: Guides
 ---
 
+{{% warn %}}
+## IMPORTANT
+Due to a known issue in InfluxDB, attempts to upgrade an InfluxDB OSS instance to
+InfluxDB Enterprise will fail.
+A fix is in place and will be released with InfluxDB v1.7.10.
+Until InfluxDB v1.7.10 is released, **DO NOT** attempt to migrate an InfluxDB OSS
+instance to InfluxDB Enterprise by following the steps in this guide.
+
+We will update this guide to reflect the new upgrade process after the release of InfluxDB 1.7.10.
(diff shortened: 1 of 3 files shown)


content/enterprise_influxdb/v1.5/guides/migration.md
```diff
@@ -9,6 +9,19 @@ menu:
     parent: Guides
 ---
 
+{{% warn %}}
+## IMPORTANT
+Due to a known issue in InfluxDB, attempts to upgrade an InfluxDB OSS instance to
+InfluxDB Enterprise will fail.
+A fix is in place and will be released with InfluxDB v1.7.10.
+Until InfluxDB v1.7.10 is released, **DO NOT** attempt to migrate an InfluxDB OSS
+instance to InfluxDB Enterprise by following the steps in this guide.
```
kelseiv: Scott, still migrate older versions? If not, perhaps we hide the guide in the meantime? ...maybe:
Currently, attempts to migrate f InfluxDB OSS to InfluxDB Enterprise fail. Please stay tuned for InfluxDB 1.7.10 which includes a fix to resolve this issue and updated migration procedures.**

sanderson: Sounds good. I'll just comment it out. The process will still be very similar, so it won't need an entire rewrite. It just doesn't work right now because of the bug in InfluxDB.

sanderson: I don't think we want to hide the guide entirely (remove it from the nav, mark as draft, etc.). There are external link linking to this doc and it's referenced in other places. So the page still exists, but it's only the warning message.

PR #2609, opened 2020-02-01
Chronograf 1.8

kelseiv:
Updates for Chronograf 1.8

Hi @glinton, @hoorayimhelping, @TCL735, @russorat please review questions in [update Chronograf 1.8 release notes](https://github.com/influxdata/docs.influxdata.com/pull/2609/commits/ea912929124c1449658ebe29efcabea090d09573)...and @russorat, will you please review [upgrade doc changes](https://github.com/influxdata/docs.influxdata.com/pull/2609/commits/eb24362149d94710654411e8df52999ef72638d9) too?
Thank you!

Hi @glinton, @russorat, please review these two commits -thank you!:
https://github.com/influxdata/docs.influxdata.com/pull/2609/commits/234aff365b195c3556b04dfbf2a536132feb6871
https://github.com/influxdata/docs.influxdata.com/pull/2609/commits/2ce4df6da459ca6f6e686142c530a8492a41283a


content/chronograf/v1.8/about_the_project/release-notes-changelog.md
```diff
@@ -1,12 +1,27 @@
...
+  chronograf_1_8:
     name: Release notes
     weight: 10
     parent: About the project
 ---
+## v1.8.0 [2020-1-31]
+
+### Features
+
+* Add query parameter to support dashboard auto refresh.
```
kelseiv: Is this true when you selected a refresh rate for a dashboard in the UI (before this feature) that the dashboard wasn't actually refreshed after the specified interval?

content/chronograf/v1.8/about_the_project/release-notes-changelog.md
```diff
@@ -1,12 +1,27 @@
...
+* Add [`etcd`](https://etcd.io/) as an alternate backend store to support multiple instances of Chronograf for high availability.
```
kelseiv: is `etcd` currently the only backend store supports multiple instance of Chronograf?

content/chronograf/v1.8/about_the_project/release-notes-changelog.md
```diff
@@ -1,12 +1,27 @@
...
+
+### Bug Fixes
+
+* Log Viewer uses a default if the mapped severity doesn't exist.
```
kelseiv: what is the default value? Can users set this default? 

also, curious if a severity, for example, crit (critcal), be "unmapped"?

content/chronograf/v1.8/about_the_project/release-notes-changelog.md
glinton: yes, but someone can easily open a pr adding another key/value store as a backend

content/chronograf/v1.8/about_the_project/release-notes-changelog.md
TCL735: No. Prior to this feature, choosing a refresh rate for the a dashboard will refresh after the specified interval. However, this selection is universal and can be changed in many places other than the dashboard page. For example, the hosts page refresh would also affect this universal refresh rate.

With this change, dashboard refresh rate is independent of any other refresh rate selection in other areas of the UI. Additionally, no actual UI interaction is required to select a refresh rate for a dashboard. Setting a url query parameter is just as valid as clicking. (This was directly in response to a user's comment that kiosk screens do not always have clicking capability.)

kelseiv: Awesome, thank you @TCL735! Is this accurate:
Add the `RefreshRate` query parameter to set the dashboard auto-refresh interval (by default, 10000 milliseconds).

kelseiv: @TCL735, found [this info](https://docs.influxdata.com/chronograf/v1.7/guides/create-a-dashboard/#configure-dashboard-wide-settings) in the docs--to be sure, should we change 15 seconds to 10000 milliseconds here, or is this a different setting?

noramullen1:
Looks good to me once the date is updated. Thanks so much Kelly!

content/chronograf/v1.7/administration/upgrading.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Upgrading to Chronograf 1.4
+title: Upgrading to Chronograf
```
sanderson: Remove "to"

content/chronograf/v1.8/about_the_project/release-notes-changelog.md
TCL735: Dashboard default refresh rate is 10000 milliseconds.

Default refresh rate elsewhere (such as Hosts Page, Data Explorer) is Paused

TCL735: Keep in mind that the refresh rate is persisted everywhere. So, the default refresh rate is only used when a refresh rate cannot be found in local storage.

content/chronograf/v1.8/about_the_project/contributing.md
```diff
@@ -0,0 +1,11 @@
...
+title: Contributing to Chronograf
+menu:
+  chronograf_1_8:
+    name: Contributing
+    weight: 20
+    parent: About the project
+    url: https://github.com/influxdata/chronograf/blob/master/CONTRIBUTING.md
+---
+
+See [Contributing to Chronograf](https://github.com/influxdata/chronograf/blob/master/CONTRIBUTING.md) in the Chronograf GitHub project to learn how you can contribute to the Chronograf project.
```
glinton: It's a bummer there's not an easily sustainable way to pull those docs in here so people aren't jumping from the docs to github

content/chronograf/v1.8/administration/configuration.md
```diff
@@ -0,0 +1,61 @@
...
+However, it is also possible to set custom default configuration options in the filesystem so they don't have to be passed in when starting Chronograf.
+
+## Starting the Chronograf service
+
+Start Chronograf using the default configuration options, or [customize your configuration](https://docs.influxdata.com/chronograf/v1.8/administration/configuration/) with environment variables and command line options (for example, to configure OAuth 2.0 authentication) based on your requirements.
+
+**Linux:**
+
+```bash
+sudo systemctl start chronograf [OPTIONS]
```
glinton: Does this actually append the options to the service? I was under the impression `systemctl` didn't take args for the process it's managing. If it's meant to match the macOS docs below to just start chronograf in the foreground, this could be the same:
```sh
chronograf [OPTIONS]
```
`sudo systemctl start chronograf` will only start the chronograf service if it has been installed (it may be installed by our packages)

content/chronograf/v1.8/administration/upgrading.md
```diff
@@ -0,0 +1,19 @@
...
+menu:
+  chronograf_1_8:
+    name: Upgrading
+    weight: 10
+    parent: Administration
+---
+
+## Upgrading to Chronograf 1.8
+
+If you're upgrading from Chronograf 1.3.x, first install version 1.4-1.7, and then install 1.8.
```
glinton: Highly unlikely, but could this be misunderstood that one has to install each of the versions from 1.4-1.7 before moving on to 1.8? Maybe be opinionated and specify `install a 1.7.x version, and...`

glinton:
There's also https://github.com/influxdata/chronograf/pull/5362 that i'm open to feedback from y'all on the choices of terminology/flags/defaults.

content/chronograf/v1.8/about_the_project/release-notes-changelog.md
hoorayimhelping: >what is the default value? Can users set this default?

the default is set in the log viewer configuration panel:

Screen Shot 2020-02-03 at 4 44 15 PM

it defaults to whatever the first one is (in this case Ruby).

>can a severity (for example, crit (critcal)) be "unmapped"?
this guards against someone entering "emreg" when they meant to type "emerg". instead of the log viewer crashing, it'll just fall back to the very first color in the list.

content/chronograf/v1.8/guides/dashboard-template-variables.md
```diff
@@ -67,6 +67,7 @@ WHERE time > :dashboardTime: AND time < :upperDashboardTime:
 ```
 
 ### interval
+
 The `:interval:` template variable is defined by the interval dropdown in the Chronograf dashboard.
 
 <img src="/img/chronograf/v1.8/template-vars-interval-dropdown.png" style="width:100%;max-width:549px;" alt="Dashboard interval selector"/>
```
kelseiv: @TCL735, @russorat, would it make sense to add the following note to this "interval" section?:
> **Note:** The :interval: template variable stores a dashboard's refresh rate and persists in local storage. If this refresh value isn't found in local storage, Chronograf uses the default refresh rate:
- for dashboards, 10 seconds
- everywhere else (Data Explorer, Host List page) refresh is paused

content/chronograf/v1.8/about_the_project/release-notes-changelog.md
kelseiv: @TCL735, I updated the dashboard settings doc and found another place that we may want to mention this feature?--in the `:interval:` template variable doc. Will stay tuned.

content/chronograf/v1.8/about_the_project/release-notes-changelog.md
kelseiv: thank you, @hoorayimhelping! Is this accurate:
* The Log Viewer no longer crashes if an invalid severity is used (for example, `emreg` instead of `emerg`). Now, by default, the Log Viewer assigns the first color on the Configure Log View page to the invalid severity.

sanderson:
@kelseiv Be sure to update `/data/default_versions.toml` and `/data/versions.toml` with the appropriate values.

content/chronograf/v1.8/about_the_project/contributing.md
kelseiv: Where there's a will....we've had some discussion around this topic and w/o an immediately apparent solution, the subject's been tabled. It'd be nice if we could mirror the files across repos.

content/chronograf/v1.8/administration/configuration.md
kelseiv: @russorat-do you know how this works?

content/chronograf/v1.8/administration/upgrading.md
kelseiv: great point. updated

kelseiv:
> There's also [influxdata/chronograf#5362](https://github.com/influxdata/chronograf/pull/5362) that i'm open to feedback from y'all on the choices of terminology/flags/defaults.

thank you, @glinton!  
> 
> Also, does this address #2610 ?
yes

content/chronograf/v1.8/administration/configuration.md
glinton: When adding flags to systemct  commands, those are processed by systemctl , not the service it's managing:
```
$ systemctl start chronograf --version
systemd 242 (242)
+PAM +AUDIT +SELINUX +IMA +APPARMOR +SMACK +SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ +LZ4 +SECCOMP +BLKID +ELFUTILS +KMOD +IDN2 -IDN +PCRE2 default-hierarchy=hybrid
```

If showing how to start the service that gets installed when installing the `.deb` package (can't confirm for `.rpm`), `[OPTIONS]` is an invalid argument. simply show how to control the service (started by default upon .deb install).
```
sudo systemctl start chronograf
```
Since there is already plenty of documentation around controlling systemd managed services, it seems out of place imo. It seems sufficient to simply show how to start chronograf manually, which is the same for each distribution (with the exception of windows, in which case it is `chronograf.exe [OPTIONS]`)

kelseiv: thanks, @glinton! 
-dropped "[OPTIONS]" from the Linux command
-added command for Windows

content/chronograf/v1.8/administration/migrate-to-chronograf-cluster.md
```diff
@@ -0,0 +1,29 @@
...
+    parent: Administration
+---
+
+Use [`chronoctl`](add link to tool) to migrate your Chronograf configuration store from BoltDB to a shared `etcd` data store used for Chronograf high-availability (HA) clusters.
+
+1. To prevent data loss, we **strongly recommend** that you back up your Chronograf data stores before migrating to a Chronograf cluster.
+
+2. [Install and start `etcd`](/chronograf/v1.8/guides/high-availability-etcd#install-and-start-etcd).
+
+3. Run the following command, specifying the local BoltDB file and the `etc` endpoint beginning with `etcd//`. (We recommend adding the prefix `bolt://` to an absolute path, but a relative path without the prefix is also accepted.)
```
kelseiv: should we drop "(We recommend.....is also accepted.) and simply say add bolt:// prefix?

content/chronograf/v1.8/administration/migrate-to-chronograf-cluster.md
```diff
@@ -0,0 +1,29 @@
...
+
+```sh
+$ chronoctl migrate -f bolt:///path/to/chronograf-v1.db -t etcd://localhost:2379
+```
+
+    > **Note:** 
+      If you have authentication on `etcd`, use the standard URI format to define a username and password. For example, `etcd://user:pass@localhost:2379 (by default, chronograf-v1.db)`
```
kelseiv: by default, chronograf-v1.db replaces "localhost:2379"?

russorat:
could we ensure this option is doc'd as well: https://github.com/influxdata/chronograf/pull/5308

content/chronograf/v1.8/about_the_project/_index.md
```diff
@@ -0,0 +1,59 @@
+---
+title: About the Chronograf project
+description: Chronograf is the UI component of the InfluxData time series platform. This section includes documentation about the Chronograf project, release notes and changelogs, what's new, contributing, and licenses.
+menu:
+  chronograf_1_8:
+    name: About the project
+    weight: 10
+---
+Chronograf is the user interface component of the [InfluxData time series platform](https://www.influxdata.com/time-series-platform/). It makes the monitoring and alerting for your infrastructure easy to setup and maintain. It is simple to use and includes templates and libraries to allow you to rapidly build dashboards with realtime visualizations of your data.
```
russorat: we aren't doing TICK stack anymore?

content/chronograf/v1.8/administration/configuration.md
```diff
@@ -0,0 +1,70 @@
...
+**Linux:**
+
+```bash
+sudo systemctl start chronograf
+```
+
+**macOS:**
+
+```bash
+chronograf [OPTIONS]
```
russorat: The option for linux starts chronograf as a service in the background, which this option starts chronograf in the foreground. Do we want to add options for both? or a note about running chronograf in the background on a mac?

content/chronograf/v1.8/administration/config-options.md
```diff
@@ -0,0 +1,437 @@
...
+
+* Linux: Setting options for develop mode and disabling reporting
+
+```sh
+  sudo systemctl start chronograf --develop --reporting-disabled
+```
+* Mac OS X: Using shortcut options to set develop mode and disable reporting
+
+```sh
+  chronograf -d -r
```
russorat: why do the examples here add `--reporting-disabled`? we can mention that option but we shouldn't direct users to start without reporting IMO.

content/chronograf/v1.8/administration/config-options.md
```diff
@@ -0,0 +1,437 @@
...
+title: Chronograf configuration options
+description: Details on configuration settings (command line options and environment variables) for Chronograf services, Kapacitor and InfluxDB connections, and OAuth 2.0 authentication providers.
+menu:
+  chronograf_1_8:
+    name: Configuration options
+    weight: 30
+    parent: Administration
+---
+
+**On this page:**
```
russorat: how is the `/etc/default/chronograf` file used? it's not clear to me when to use it vs setting things on the command line.

content/chronograf/v1.8/administration/configuration.md
```diff
@@ -0,0 +1,70 @@
...
+
+```conf
+HOST=0.0.0.0
+PORT=8888
+TLS_CERTIFICATE=/path/to/cert.pem
+TOKEN_SECRET=MySup3rS3cretT0k3n
+LOG_LEVEL=info
+```
+
+> **Note:** `/etc/default/chronograf` is only created in Linux-based operating systems.
```
russorat: can you use it on other systems? or only on Linux?

content/chronograf/v1.8/administration/creating-connections.md
```diff
@@ -0,0 +1,238 @@
...
+---
+
+Connections to InfluxDB and Kapacitor can be configured through the Chronograf user interface (UI) or with JSON configuration files:
+
+- [Manage InfluxDB connections using the Chronograf UI](#manage-influxdb-connections-using-the-chronograf-ui)
+- [Manage InfluxDB connections using .src files](#manage-influxdb-connections-using-src-files)
+- [Manage Kapacitor connections using the Chronograf UI](#manage-kapacitor-connections-using-the-chronograf-ui)
+- [Manage Kapacitor connections using .kap files](#manage-kapacitor-connections-using-kap-files)
+
+> Note: Connection details are stored in Chronograf’s internal database `chronograf-v1.db` `.src`. Typically, the only time you'll need to administer the internal database is when you're [restoring a Chronograf database](/chronograf/v1.8/administration/restoring-chronograf-db/).
```
russorat: the database can be boltdb or etcd now, so this example might need to be updated.

content/chronograf/v1.8/administration/creating-connections.md
```diff
@@ -0,0 +1,238 @@
...
+
+    * **Connection URL**: Enter the hostname or IP address of the InfluxDB instance and the port. The field is prefilled with  `http://localhost:8086`.
+    * **Connection Name**: Enter the name for this connection.
+    * **Username**: Enter the username that will be shared for this connection.
+      *Required only if [authorization is enabled](/influxdb/latest/administration/authentication_and_authorization/) on the InfluxDB instance to which you're connecting.*
+    * **Password**: Enter the password.
+      *Required only if [authorization is enabled](/influxdb/latest/administration/authentication_and_authorization/) on the InfluxDB instance to which you're connecting.*
+    * **Telegraf Database Name**: This field specifies the database that Chronograf will use for populating different parts of the application, including the Host List page. If you've changed the default retention policy in your InfluxDB instance, you may want to change it here as well. If none is provided, we will use `autogen`. You will still be able to query any database you have access to in the InfluxDB instance when building dashboards or exploring data.
```
russorat: The second sentence about the rp should go in the next bullet IMO.

content/chronograf/v1.8/administration/creating-connections.md
```diff
@@ -0,0 +1,238 @@
...
+| InfluxDB Enterprise   | `influx-enterprise` |
+
+#### `insecureSkipVerify`  
+Skips the SSL certificate verification process.
+Set to `true` if you are using a self-signed SSL certificate on your InfluxDB server or cluster.
+
+#### `default`  
+Set to `true` if you want the connection to be the default data connection used upon first login.
+
+#### `telegraf`
```
russorat: is there a way to set rp in this file?

content/chronograf/v1.8/administration/migrate-to-chronograf-cluster.md
```diff
@@ -0,0 +1,29 @@
+---
+title: Migrate to a Chronograf cluster
+description: Migrate to a Chronograf high-availability (HA) cluster.
+menu:
+  chronograf_1_8:
+    weight: 10
+    parent: Administration
```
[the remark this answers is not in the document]
russorat: Is "cluster" the right word here? I would call this "Migrating to a Highly Available Chronograf Configuration" or something. Cluster gives the impression that each chronograf instance communicates to the other instances, which isn't the case.

content/chronograf/v1.8/administration/migrate-to-chronograf-cluster.md
```diff
@@ -0,0 +1,29 @@
...
+---
+
+Use [`chronoctl`](add link to tool) to migrate your Chronograf configuration store from BoltDB to a shared `etcd` data store used for Chronograf high-availability (HA) clusters.
+
+1. To prevent data loss, we **strongly recommend** that you back up your Chronograf data stores before migrating to a Chronograf cluster.
+
+2. [Install and start `etcd`](/chronograf/v1.8/guides/high-availability-etcd#install-and-start-etcd).
```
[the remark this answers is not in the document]
russorat: yes, it needs to be started.

content/chronograf/v1.8/administration/migrate-to-chronograf-cluster.md
russorat: `etcd//` should be `etcd://`?

content/chronograf/v1.8/administration/migrate-to-chronograf-cluster.md
```diff
@@ -0,0 +1,29 @@
...
+description: Migrate to a high-availability (HA) Chronograf cluster.
+menu:
+  chronograf_1_8:
+    weight: 10
+    parent: Administration
+---
+
+Use [`chronoctl`](add link to tool) to migrate your Chronograf configuration store from BoltDB to a shared `etcd` data store used for Chronograf high-availability (HA) clusters.
+
+1. To prevent data loss, we **strongly recommend** that you back up your Chronograf data stores before migrating to a Chronograf cluster.
```
sanderson: "stores" should be singular. Chronograf only has one data store.

content/chronograf/v1.8/administration/migrate-to-chronograf-cluster.md
sanderson: @glinton Does the migration process completely overwrite whatever is in the destination datastore?

content/chronograf/v1.8/administration/migrate-to-chronograf-cluster.md
```diff
@@ -0,0 +1,29 @@
...
+```
+
+    > **Note:** 
+      If you have authentication on `etcd`, use the standard URI format to define a username and password. For example, `etcd://user:pass@localhost:2379 (by default, chronograf-v1.db)`
+
+4. Update links to Chronograf (for example, from external sources) to reflect your new URLs:
+    - **from BoltDB:**
+    http://localhost:8888/sources/1/status
+    - **to etcd:**
+    http://localhost:8888/sources/373921399246786560/status
```
sanderson: We may need to explain why they need to update these.

> The migration process requires updates to all Chronograf source URLs...

content/chronograf/v1.8/administration/migrate-to-chronograf-cluster.md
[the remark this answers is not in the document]
sanderson: I think you have a point @russorat, but I'm curious if "cluster" is a search term that will bring people here.

content/chronograf/v1.8/administration/configuration.md
glinton: i believe it's created when installing the `.deb` or `.rpm` package

content/chronograf/v1.8/administration/migrate-to-chronograf-cluster.md
glinton: it adds to it, so in that way it is not idempotent

sanderson: @kelseiv This is an important detail to mention ☝️

russorat: i don't think this default applies here.

content/chronograf/v1.8/administration/migrate-to-chronograf-cluster.md
glinton: it should be `etcd://` and '`etc`  endpoint' should be '`etcd` endpoint'

content/chronograf/v1.8/administration/migrate-to-chronograf-cluster.md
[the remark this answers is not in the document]
glinton: it definitely needs started, installing the chronograf package (`.deb`/`.rpm`) tries to automatically start the chronograf service, but since it isn't guaranteed to be running after install. mentioning it should be started lets users know to double check that it is running.

content/chronograf/v1.8/administration/migrate-to-chronograf-cluster.md
```diff
@@ -0,0 +1,29 @@
...
+3. Run the following command, specifying the local BoltDB file and the `etc` endpoint beginning with `etcd//`. (We recommend adding the prefix `bolt://` to an absolute path, but a relative path without the prefix is also accepted.)
+
+```sh
+$ chronoctl migrate -f bolt:///path/to/chronograf-v1.db -t etcd://localhost:2379
+```
+
+    > **Note:** 
+      If you have authentication on `etcd`, use the standard URI format to define a username and password. For example, `etcd://user:pass@localhost:2379 (by default, chronograf-v1.db)`
+
+4. Update links to Chronograf (for example, from external sources) to reflect your new URLs:
```
russorat: I think it would be worthwhile to add a note at the beginning of the migration process that the source ids will change and that you will need to update any external links to dashboards.

content/chronograf/v1.8/administration/prebuilt-dashboards.md
```diff
@@ -0,0 +1,486 @@
...
+description: Prebuilt dashboards available to import based on Telegraf input plugins.
+menu:
+  chronograf_1_8:
+    name: Prebuilt dashboards in Chronograf
+    weight: 50
+    parent: Administration
+---
+Chronograf lets you import a variety of prebuilt dashboards that visualize metrics collect by specific [Telegraf input plugins](/telegraf/latest/plugins/plugin-list). The following Telegraf-related dashboards templates are available.
+
+For details on how to import dashboards while adding a connection in Chronograf, see [Creating connections](/chronograf/v1.8/administration/creating-connections//#managing-influxdb-connections-using-the-chronograf-ui).
```
russorat: extra / in the url

content/chronograf/v1.8/administration/restoring-chronograf-db.md
```diff
@@ -0,0 +1,81 @@
...
+
+Chronograf's internal database, `chronograf-v1.db`, is stored at your specified
+[`--bolt-path`](/chronograf/v1.8/administration/config-options/#bolt-path-b) which,
+by default, is the current working directory where the `chronograf` binary is executed.
+In the upgrade process, an unmodified backup of your Chronograf data is stored inside the
+`backup` directory before any necessary migrations are run.
+This is done as a convenience in case issues arise with the data migrations
+or the upgrade process in general.
+
+The `backup` directory is a sibling of your `chronograf-v1.db` file.
```
russorat: i wouldn't call it a sibling. The backup directory contains a copy of your previous boltdb file when you change versions of chronograf. It isn't touched after the initial backup when telegraf is started.

content/chronograf/v1.8/administration/migrate-to-chronograf-cluster.md
```diff
@@ -0,0 +1,29 @@
...
+description: Migrate to a Chronograf high-availability (HA) cluster.
+menu:
+  chronograf_1_8:
+    weight: 10
+    parent: Administration
+---
+
+Use [`chronoctl`](add link to tool) to migrate your Chronograf configuration store from BoltDB to a shared `etcd` data store used for Chronograf high-availability (HA) clusters.
+
+1. To prevent data loss, we **strongly recommend** that you back up your Chronograf data stores before migrating to a Chronograf cluster.
```
russorat: you need to stop your chronograf instance prior to running this command.

content/chronograf/v1.8/administration/creating-connections.md
glinton: Also, the addition of `.src` seems like a typo. `.src` files placed in chronograf's `--resources-path` will expose those sources by default on chronograf's start.

content/chronograf/v1.8/guides/create-a-dashboard.md
```diff
@@ -0,0 +1,130 @@
...
+Click the **Add a Query** button to create an [InfluxQL](/influxdb/latest/query_language/) query.
+In query editor mode, use the builder to select from your existing data and allow Chronograf to format the query for you.
+Alternatively, manually enter and edit a query.
+Chronograf allows you to move seamlessly between using the builder and manually editing the query; when possible, the interface automatically populates the builder with the information from your raw query.
+
+For our example, the query builder is used to generate a query that shows the average idle CPU usage grouped by host (in this case, there are three hosts).
+By default, Chronograf applies the [`MEAN()` function](/influxdb/latest/query_language/functions/#mean) to the data, groups averages into auto-generated time intervals (`:interval:`), and shows data for the past hour (`:dashboardTime:`).
+Those defaults are configurable using the query builder or by manually editing the query.
+
+In addition, the time range (`:dashboardTime:`) is [configurable on the dashboard](#step-6-configure-your-dashboard).
```
russorat: i think this might now also have an `:upperDashboardTime:` as well.

content/chronograf/v1.8/guides/dashboard-template-variables.md
```diff
@@ -0,0 +1,399 @@
...
+
+`:database:`  
+`:measurement:`  
+`:dashboardTime:`  
+`:upperDashboardTime:`  
+`:interval:`  
+`:upper:`  
+`:lower:`  
+`:zoomedUpper:`  
+`:zoomedLower:`
```
russorat: there is now one for refreashRate i think: https://github.com/influxdata/chronograf/pull/5348

content/chronograf/v1.8/guides/high-availability-etcd.md
```diff
@@ -0,0 +1,32 @@
+---
+title: Creating a Chronograf cluster
+description: Create a Chronograf high-availability (HA) cluster using etcd.
+menu:
+  chronograf_1_8:
+    weight: 100
+    parent: Guides
+---
+
+To create a Chronograf high-availability (HA) cluster using etcd as a shared data store, do the following:
```
russorat: again, i'm not sure "cluster" is the right word here since chronografs don't talk to eachother.

content/chronograf/v1.8/administration/migrate-to-chronograf-cluster.md
glinton: :+1:  The reason it is different is because of how the id's are generated. Technically, it was for backwards compatibility. Dashboard id's were stored and transmitted as integers rather than strings. We implemented a universally unique integer generator for etcd (boltdb has a global counter internally that was used, hence the typically low number id's). These uuid's are large numbers, hence the difference. We could have preserved the id's upon migration, but felt that it would be "safer" to generate new id's on migrate.

content/chronograf/v1.8/guides/monitoring-influxenterprise-clusters.md
```diff
@@ -0,0 +1,263 @@
...
+  [...]
+
+  ## Write timeout (for the InfluxDB client), formatted as a string.
+  ## If not provided, will default to 5s. 0s means no timeout (not recommended).
+  timeout = "5s"
+  username = "chronothan" #💥
+  password = "supersecret" #💥
+```
+
+The [Telegraf System input plugin](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/system) is enabled by default and requires no additional configuration.
```
russorat: The cpu, disk, etc have been split out into separate plugins. not sure if you want to call that out here or not.

content/chronograf/v1.8/guides/monitoring-influxenterprise-clusters.md
```diff
@@ -0,0 +1,263 @@
...
+  totalcpu = true
+  ## If true, collect raw CPU time metrics.
+  collect_cpu_time = false
+```
+
+#### Step 3: Restart the Telegraf service.
+
+Restart the Telegraf service so that your configuration changes take effect:
+```
+~# sudo systemctl restart telegraf
```
russorat: do we need examples from different platforms here?

content/chronograf/v1.8/guides/monitoring-influxenterprise-clusters.md
```diff
@@ -0,0 +1,263 @@
...
+```
+
+Repeat steps one through four for each data node in your cluster.
+
+#### Step 4: Confirm the Telegraf setup.
+
+Run the following command on your InfluxDB OSS instance to see if your Telegraf instances are successfully collecting and writing data.
+Replace the `chronothan` and `supersecret` values with your actual username and password.
+```
+~# curl -G "http://localhost:8086/query?db=telegraf&u=chronothan&p=supersecret&pretty=true" --data-urlencode "q=SHOW TAG VALUES FROM cpu WITH KEY=host"
```
russorat: why not use the CLI?

content/chronograf/v1.8/guides/using-precreated-dashboards.md
```diff
@@ -0,0 +1,476 @@
...
+3. The selected application appears in a page showing all of the pre-created cells, based on the measurements that are available.
+
+## Creating or editing dashboards
+
+The pre-created dashboards available for use with Chronograf are listed below as Apps and appear in the Apps listing for hosts in the Host List page.  The listings for apps (pre-created dashboards) on this page include information about the required Telegraf input plugins, JSON files included in the apps, and listings of the cell titles that are included in each of the JSON files.
+
+The JSON files for the precreated dashboards are delivered on installation in the `/usr/share/chronograf/canned` directory. Information about the  configuration option `--canned-path` can be found in the [Chronograf configuration options](/chronograf/latest/administration/config-options/#) page.
+
+You can enable and disable applications in your [Telegraf configuration file](https://github.com/influxdata/telegraf/blob/master/etc/telegraf.conf).
+See [Telegraf configuration](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md) for details.
```
russorat: do we have a docs.influxdata version of this?

content/chronograf/v1.8/administration/migrate-to-chronograf-cluster.md
glinton: https://github.com/influxdata/chronograf/pull/5377

content/chronograf/v1.8/administration/migrate-to-chronograf-cluster.md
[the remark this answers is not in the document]
kelseiv: thanks @russorat, @sanderson  - do you guys see “migrating the configuration” as part of “Creating a Chronograf HA cluster” process?

...or if we say “Migrating to a Highly Available Chronograf Configuration,” does it also make sense to change “Creating a Chronograf HA cluster” to  “Creating a Chronograf HA configuration”? 

re: terms folks search for, we can leave "cluster" in the description, and use the most precise terminology/process in the heading

content/chronograf/v1.8/about_the_project/_index.md
kelseiv: @russorat, is there something in the content that makes you ask this question?
Content rolled over from 1.7...let me know if we need to change something

content/chronograf/v1.8/administration/configuration.md
kelseiv: Is starting chronograf in the background or foreground preferable/more common? Ideally, we'd give the same option for both Linux and Mac, and then add a note for the other. 
What is the command to start chronograf in the foreground on linux/in the background on macOS?

content/chronograf/v1.8/administration/config-options.md
kelseiv: reworded to:
**Linux examples**

- To start `chronograf` without options:

```sh
  sudo systemctl start chronograf
```

- To start `chronograf` and set options for develop mode and to disable reporting:

```sh
  sudo systemctl start chronograf --develop --reporting-disabled
```

**MacOS X examples**

- To start `chronograf` without options:

```sh
  chronograf
```

To start `chronograf` and add shortcut options for develop mode and to disable reporting:

```sh
  chronograf -d -r
```

content/chronograf/v1.8/administration/config-options.md
kelseiv: Good catch, @russorat. I added a statement similar to what's said about InfluxDB as an intro paragraph. Lmk if this addresses:

Chronograf is configured using the configuration file (/etc/default/chronograf) and environment variables. If you do not uncomment a configuration option, the system uses its default setting. The configuration settings below are set to their default settings. For more information, see [Configuration Chronograf](/chronograf/v1.8/administration/configuration/).

content/chronograf/v1.8/administration/configuration.md
kelseiv: good question, this note has been there for awhile.... @glinton @hoorayimhelping, @sanderson- do any of you know the answer?

content/chronograf/v1.8/administration/creating-connections.md
kelseiv: @russorat, reworded this note to say:

Connection details are stored in Chronograf’s internal database `chronograf-v1.db` `.src`. You may administer the internal database when [restoring a Chronograf database](/chronograf/v1.8/administration/restoring-chronograf-db/) or [migrating from BoltDB to etcd](link header TBD).

Although, I'm not sure whether migrating the configuration is technically administering the `chronograf-v1.db`?

content/chronograf/v1.8/administration/creating-connections.md
kelseiv: Moved this sentence to the next section:
"If you've changed the default retention policy in your InfluxDB instance, you may want to change it here as well."

content/chronograf/v1.8/administration/creating-connections.md
kelseiv: @glinton @hoorayimhelping @sanderson?

content/chronograf/v1.8/administration/migrate-to-chronograf-cluster.md
kelseiv: added a note at the beginning of the procedure

content/chronograf/v1.8/administration/migrate-to-chronograf-cluster.md
kelseiv: added:
> **Note:**  Migrating your Chronograf data source changes the source IDs, so you must update external links to Chronograf dashboards to reflect the new source IDs.

content/chronograf/v1.8/administration/restoring-chronograf-db.md
kelseiv: Is this accurate:
The `backup` directory contains a copy of your previous `chronograf-v1.db` file.

content/chronograf/v1.8/administration/migrate-to-chronograf-cluster.md
kelseiv: added as step 1

content/chronograf/v1.8/guides/create-a-dashboard.md
kelseiv: updated:
In addition, the time range (`:dashboardTime:` and `:upperDashboardTime:`) are [configurable on the dashboard](#step-6-configure-your-dashboard).

content/chronograf/v1.8/guides/dashboard-template-variables.md
kelseiv: added

content/chronograf/v1.8/guides/monitoring-influxenterprise-clusters.md
kelseiv: Updated text to reflect plugin currently collects system load, uptime, and number of users logged in

content/chronograf/v1.8/administration/configuration.md
kelseiv: updated to say:
> **Note:** `/etc/default/chronograf` is only created when installing the `.deb or `.rpm` package.

content/chronograf/v1.8/administration/creating-connections.md
hoorayimhelping: what is rp?

kelseiv: @hoorayimhelping, think @russorat is referring to retention policy?

content/chronograf/v1.8/administration/migrate-to-chronograf-cluster.md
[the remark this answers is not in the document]
kelseiv: @russorat, @sanderson, updated filenames, titles, and descriptions to:

- filename: create-high-availability-etcd.md
- title: Creating a Chronograf HA configuration
- description: Create a Chronograf high-availability (HA) cluster using etcd.

- filename: migrate-to-high-availability-etcd.md
- title: Migrating to a Chronograf HA configuration
- description: Migrate a Chronograf single instance configuration using BoltDB to a Chronograf high-availability (HA) cluster configuration using etcd.

content/chronograf/v1.8/about_the_project/release-notes-changelog.md
TCL735: Yes, we should mention that `:interval:` is essentially the "group by" in the query, but _not_ the refresh rate. Refresh rate is now tied only to the dropdown and is the url parameter called **refresh**. In other words `:interval:` and refresh rate are not connected.

content/chronograf/v1.8/administration/configuration.md
russorat: I think these should work for both mac and linux: https://github.com/influxdata/chronograf#installation

content/chronograf/v1.8/administration/creating-connections.md
russorat: yes retention policy

content/chronograf/v1.8/administration/configuration.md
kelseiv: thank you, @russorat!.... updated the content in _Configuring Chronograf_.

content/chronograf/v1.8/guides/monitoring-influxenterprise-clusters.md
kelseiv: thanks, @russorat - added examples of restarting on different platforms

content/chronograf/v1.8/guides/monitoring-influxenterprise-clusters.md
kelseiv: Good question, @russorat. Probably a global issue across the 1.x docs. Typically, when users can complete a task in multiple ways, you'd tell them one way to do it in task-based doc for simplicity (and use the same method across all task-based doc), and then delve into all options in reference doc. I'll discuss this issue w the docs-team tomorrow.

content/chronograf/v1.8/guides/using-precreated-dashboards.md
kelseiv: yes, updated reference - thank you, @russorat

content/chronograf/v1.8/administration/migrate-to-chronograf-cluster.md
kelseiv: updated:
For example, `etcd://user:pass@localhost:2379`

content/chronograf/v1.8/administration/creating-connections.md
kelseiv: @hoorayimhelping, do you know if folks can set the retention policy in the 
Chronograf config file or is that something that must be done in influxdb config?

content/chronograf/v1.8/guides/monitoring-influxenterprise-clusters.md
kelseiv: @russorat, added example for CLI.

content/chronograf/v1.8/guides/using-precreated-dashboards.md
kelseiv: @russorat, do you know how to "enable and disable applications" in the Telegraf configuration file? After reading the conf file on both github and the doc.influxdata info, it's not clear to me.

I found the following reference to "app" in the Telegraf .conf file- is this related?:

# Input plugin for DC/OS metrics
[[inputs.dcos]]
....
Container IDs to collect app metrics from.
app_include = []
app_exclude = []

content/chronograf/v1.8/about_the_project/release-notes-changelog.md
hoorayimhelping: @kelseiv yep, looks good!

content/chronograf/v1.8/administration/create-high-availability-etcd.md
```diff
@@ -0,0 +1,37 @@
...
+   (For detailed installation instructions specific to your operating system, see [Install and deploy etcd](http://play.etcd.io/install).)
+2. Extract the `etcd` binary and place it in your system PATH.
+3. Start etcd.
+
+## Start Chronograf
+
+Run the following command to start Chronograf using etcd as the storage layer:
+
+```sh
+chronograf --etcd-endpoints=localhost:2379 --port <insert-port-number>
```
russorat: --port is optional. i think it can be removed here.

content/chronograf/v1.8/administration/creating-connections.md
russorat: We can follow up on this later in a separate issue IMO.

content/chronograf/v1.8/administration/create-high-availability-etcd.md
```diff
@@ -0,0 +1,37 @@
+---
```
sanderson: I'd suggest renaming this file to `chronograf-high-availability.md` or just `high-availability.md`. We'll also need to update any links to the old URL.

content/chronograf/v1.8/administration/create-high-availability-etcd.md
```diff
@@ -0,0 +1,37 @@
...
+2. Extract the `etcd` binary and place it in your system PATH.
+3. Start etcd.
+
+## Start Chronograf
+
+Run the following command to start Chronograf using etcd as the storage layer:
+
+```sh
+chronograf --etcd-endpoints=localhost:2379 --port <insert-port-number>
+```
```
sanderson:
~~~
```sh
# Sytnax
chronograf --etcd-endpoints=<etcd-host>

# Example
chronograf --etcd-endpoints=localhost:2379
```
~~~

content/chronograf/v1.8/administration/migrate-to-high-availability-etcd.md
```diff
@@ -0,0 +1,32 @@
+---
```
sanderson: Let's rename this file with a shorter name – `migrate-to-ha`, `migrate-to-high-availability`. We'll need to update any links to this doc.

content/chronograf/v1.8/administration/migrate-to-high-availability-etcd.md
```diff
@@ -0,0 +1,32 @@
...
+title: Migrating to a Chronograf HA configuration
+description: Migrate a Chronograf single instance configuration using BoltDB to a Chronograf high-availability (HA) cluster configuration using etcd.
```
sanderson: Switch to imperative voice.

- Migrate to...
- Migrate a...

content/chronograf/v1.8/administration/migrate-to-high-availability-etcd.md
```diff
@@ -0,0 +1,32 @@
...
+menu:
+  chronograf_1_8:
+    weight: 10
+    parent: Administration
+---
+
+Use [`chronoctl`](/chronograf/v1.8/tools/chronoctl/) to migrate your Chronograf configuration store from BoltDB to a shared `etcd` data store used for Chronograf high-availability (HA) clusters.
+
+> **Note:**  Migrating your Chronograf to a shared data source creates new source IDs for each resource. You must update external links to Chronograf dashboards to reflect new source IDs.
```
sanderson: **Note:**  Migrating ~your~ Chronograf to a shared data source creates new source IDs for each resource. ~You must~ Update external links to Chronograf dashboards to reflect new source IDs.

content/chronograf/v1.8/administration/migrate-to-high-availability-etcd.md
```diff
@@ -0,0 +1,32 @@
...
+
+1. Stop the Chronograf server by killing the `chronograf` process.
+2. To prevent data loss, we **strongly recommend** that you back up your Chronograf data store before migrating to a Chronograf cluster.
```
sanderson: @russorat What's the recommended way to backup Chronograf data?

content/chronograf/v1.8/administration/transition-web-admin-interface.md
```diff
@@ -0,0 +1,94 @@
+---
```
sanderson: I think we can remove this doc in the Chronograf 1.8 docs.

content/chronograf/v1.8/administration/create-high-availability-etcd.md
kelseiv: @sanderson, looks like most of the procedures currently use task-based filenames....
If we rename this file, what to name the migrate-to-high-availability-etcd file?

content/chronograf/v1.8/administration/migrate-to-high-availability-etcd.md
kelseiv: I used the gerund in the title intentionally to be consistent...is that what you're referiring to?

I agree, we could update this throughout the doc set

content/chronograf/v1.8/administration/create-high-availability-etcd.md
sanderson: I'm not too concerned about having task-based file names. File names determine the URL of the doc and it's generally good practice to keep URLs shorter when possible. Looking into possible changes in the future, we shouldn't include `etcd` in the filename. It's very possible that support for other backends will be introduced.

kelseiv: changed to create-high-availability

content/chronograf/v1.8/administration/migrate-to-high-availability-etcd.md
sanderson: Yeah, that's what I'm referring to. This may be the wrong approach because it does create inconsistency among doc titles, but for anything new, I've been using imperative voice in titles. We probably should go back and update other doc titles.

As a side note, minor releases are a good opportunity to do this. Should we take the time today to update doc titles in the Chronograf 1.8 docs?

kelseiv: @sanderson, added issue to update:
https://github.com/influxdata/docs.influxdata.com/issues/2659

content/chronograf/v1.8/administration/transition-web-admin-interface.md
kelseiv: removed

PR #2629, opened 2020-02-05
 update 'series-id-set-cache-size' description

kelseiv:
update description for *series-id-set-cache-size*

### DIFF
--- a/content/enterprise_influxdb/v1.7/administration/config-data-nodes.md
+++ b/content/enterprise_influxdb/v1.7/administration/config-data-nodes.md
@@ -327,14 +327,10 @@ Environment variable: `INFLUXDB_DATA_MAX_INDEX_LOG_FILE_SIZE`
 
 #### `series-id-set-cache-size = 100`
 
-The size of the internal cache used in the TSI index to store previously
-calculated series results. Cached results will be returned quickly from the cache rather
-than needing to be recalculated when a subsequent query with a matching tag key-value
-predicate is executed.
-Setting this value to `0` will disable the cache, which may lead to query performance issues.
-This value should only be increased if it is known that the set of regularly used
-tag key-value predicates across all measurements for a database is larger than 100. An
-increase in cache size may lead to an increase in heap usage.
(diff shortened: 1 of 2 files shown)


content/enterprise_influxdb/v1.7/administration/config-data-nodes.md
```diff
@@ -327,14 +327,10 @@ Environment variable: `INFLUXDB_DATA_MAX_INDEX_LOG_FILE_SIZE`
...
-calculated series results. Cached results will be returned quickly from the cache rather
-than needing to be recalculated when a subsequent query with a matching tag key-value
-predicate is executed.
-Setting this value to `0` will disable the cache, which may lead to query performance issues.
-This value should only be increased if it is known that the set of regularly used
-tag key-value predicates across all measurements for a database is larger than 100. An
-increase in cache size may lead to an increase in heap usage.
+Specifies the number of series ID sets to cache for the TSI index (by default, 100). The internal LRU cache removes the least recently used series ID sets and caches only the specified number for easy access in memory. Cached results are returned quickly because they don’t need to be recalculated when a subsequent query with a matching tag key-value predicate is executed. For example, if a query includes `region = west`, the series IDs matching `region = west` are cached and subsequent queries that include `region = west` are retrieved from memory rather than disk. Each series ID set can store up to 512MB (with 4B series).
```
e-dard: I think perhaps this could be rephrased slightly:

```
Specifies the number of series ID sets to cache for the TSI index (by default, 100). All series IDs in a set refer to series that all match on the same index predicate (tag pair filter). An example filter might be `region = west`. When the index plans a query it produces one of these sets for each tag filter in the query. These sets are then cached in the index.

THe series ID set is an LRU cache, so once it is full the least recently used set is evicted. Cached results are returned quickly because they don’t need to be recalculated when a subsequent query with a matching tag filter is executed. For example, if a query includes `region = west`, the series IDs matching `region = west` are cached and subsequent queries that include `region = west` are retrieved from the cache.
```

PR #2638, opened 2020-02-07
Chronograf 1.8 select data source for template variable

noramullen1:


content/chronograf/v1.8/guides/dashboard-template-variables.md
```diff
@@ -88,14 +88,14 @@ Chronograf lets you create custom template variables powered by meta queries or
...
 
 1. Click on **Template Variables** at the top of your dashboard, then **+ Add Variable**.
-2. Provide a name for the variable.
-3. Select the [variable type](#template-variable-types).
+2. Select a data source from the **Data Source** dropdown menu.
+3. Provide a name for the variable.
+4. Select the [variable type](#template-variable-types).
    The type defines the method for retrieving the array of possible values.
-4. View the list of potential values and select a default.
+5. View the list of potential values and select a default.
```
hoorayimhelping: Maybe nitpicky / pedantic:

> View the list of potential values and select a **default.**

I think the context has changed and default may not be the best word now. Previously you'd select the default, but now you're actually choosing the active (or selected or current) data source for the template variable, rather than the default.

noramullen1: @hoorayimhelping This step isn't actually referring to the data source option--that's in step 2 above. (See, the real numbers are handy! 😄  ) This is referring to the setting a default from the meta query returned values, the last step in the template variable editing overlay window.

hoorayimhelping: ahhh okay, thanks :)

hoorayimhelping: >(See, the real numbers are handy! 😄 ) 

har har har 😝

PR #2642, opened 2020-02-10
Add Chronograf CLI documentation

noramullen1:
@sanderson Should we delete the old version of this page now? https://docs.influxdata.com/chronograf/v1.7/administration/config-options/


content/chronograf/v1.8/tools/_index.md
```diff
@@ -1,11 +1,18 @@
 ---
 title: Chronograf Tools
-description: >
- placeholder
+description:
```
sanderson: Provide a description here so we can control what search engines pull. Even if it's the same as the article intro.

content/chronograf/v1.8/tools/chronograf-cli/_index.md
```diff
@@ -0,0 +1,122 @@
+---
+title: chronograf cli
+description:
```
sanderson: Same as above

sanderson:
@noramullen1 About deleting the old config options doc – I'm fine with it as long as we aren't missing any key pieces of information that are in that doc. Tables don't make it easy to give more detailed information. If we delete the config-options page, we need to be sure to add it as an alias to the new CLI page so links that point to it will get redirected to the right place.

Issue #2621, opened 2020-02-05
Documentation around Windows not an officially supported platform?


sjwang90:
Support Engineer reached out that they often get Windows tickets and can't do much with them because we do not fully support Windows since it's considered ["experimental"](https://www.influxdata.com/blog/running-the-tick-stack-on-windows/) for everything besides Telegraf. 

Suggestion was to indicate somewhere on the downloads/docs page that Windows isn't an official supported platform - could suggest to use docker container on Windows. 

Open for discussion or curious if there has been conversation of this in the past.

kelseiv:
Hi @sjwang90, yes this issue has a past. Verified with @timhallinflux @influxdata/docs-team our position remains Windows's experimental, and we'll review docs to be sure we're consistent in messaging and add note on the main docs page.

PR #2645, opened 2020-02-12
add experimental Windows support to home page

kelseiv:
Include Windows is experimental on home page per https://github.com/influxdata/docs.influxdata.com/issues/2621

### DIFF
--- a/layouts/index.html
+++ b/layouts/index.html
@@ -6,7 +6,7 @@
 	<div class="homepage--card card-platform">
 		<div class="homepage--card-container">
 			<h1 class="homepage--card-title">InfluxData Platform</h1>
-			<p class="homepage--card-description">The leading time series platform built for metrics and events.</p>
+			<p class="homepage--card-description">The leading time series platform built for metrics and events. Fully supported on multiple operating systems, including Linux, MacOS, and Unix. Windows not fully supported and provided for experimental use only.</p>
 			<a class="btn" href="/platform/introduction">Start Here</a>
 		</div>
 		<div class="homepage--card-mask"></div>
--- a/layouts/index.html
+++ b/layouts/index.html
@@ -6,7 +6,7 @@
 	<div class="homepage--card card-platform">
 		<div class="homepage--card-container">
 			<h1 class="homepage--card-title">InfluxData Platform</h1>
-			<p class="homepage--card-description">The leading time series platform built for metrics and events. Fully supported on multiple operating systems, including Linux, MacOS, and Unix. Windows not fully supported and provided for experimental use only.</p>
+			<p class="homepage--card-description">The leading time series platform built for metrics and events.</p>
 			<a class="btn" href="/platform/introduction">Start Here</a>
 		</div>
 		<div class="homepage--card-mask"></div>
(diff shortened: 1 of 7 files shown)


layouts/index.html
```diff
@@ -6,7 +6,7 @@
 	<div class="homepage--card card-platform">
 		<div class="homepage--card-container">
 			<h1 class="homepage--card-title">InfluxData Platform</h1>
-			<p class="homepage--card-description">The leading time series platform built for metrics and events.</p>
+			<p class="homepage--card-description">The leading time series platform built for metrics and events. Fully supported on multiple operating systems, including Linux, MacOS, and Unix. Windows not fully supported and provided for experimental use only.</p>
```
sanderson: - MacOS -> macOS (this is the trademarked form of the name)
- Windows support is experimental.

kelseiv: Thanks, @sanderson. Will add to install docs for sure. Wanted to get input on text first. While the request was originally to include on doc homepage and/or download page itself, I'm open to removing it from docs home page. It does get the fact Windows is experimental out in front.

kelseiv: Thanks, @sanderson. Will add to install docs for sure. Wanted to get input on text first. While the request was originally to include on doc homepage and/or download page itself, I'm open to removing it from docs home page. It does get the fact Windows is experimental out in front.

kelseiv: Thanks, @sanderson. Will add to install docs for sure. Wanted to get input on text first. While the request was originally to include on doc home page and/or download page itself, I'm open to removing it from docs home page. It does get the fact Windows is experimental out in front.

sanderson: It does, but I don't know that it's apparent enough. It get's buried in a wall of images/text and I don't know how many people will actually read it.

kelseiv: @sanderson, very true. Think we all agree on install docs note.

@sjwang90, @sanderson, @noramullen1, @pierwill ...thoughts on possible other options?:
-  leave info off our main docs page
- note on main docs page under the Start Here button 
- note on download portal
- mention in install docs is all we need

kelseiv: Regarding the text  "Windows support is experimental" 

Could this be interpreted:
- We don't support Windows, but you can experiment 
OR
- We’re striving to support Windows but still in experimental phase/testing 

Perhaps we're not sure yet? Want to set expectations on what to expect for support.

sanderson: I think both of those interpretations are true.

sjwang90: I think these two would be the best options:
- note on download portal (only Under Windows downloads links)
- mention in install docs is all we need

Don't need it on any main pages, wouldn't want to detract any users.  

I agree on both the interpretations.

kelseiv: Thanks, @sanderson @sjwang90, updated all install docs and slacked @ChrisChurilo to see if she can help us add text "Windows support is experimental"
to download portal (only Under Windows downloads links)

kelseiv: per @lukevmorris, download portal issues can be submitted to [influxdb-customers repo](https://github.com/influxdata/influxdb-customers). Thank you, @sjwang90 for submitting another issue on that repo--closing this one.

PR #2660, opened 2020-02-19
fix refresh parameter name

TCL735:
`refresh` is the actual name of the parameter for the dashboard refresh rate


kelseiv:
> `refresh` is the actual name of the parameter for the dashboard refresh rate

Thank you, @TCL735! Pushed update up to docs.influxdata.com.

Issue #2668, opened 2020-02-25
What's max index log file size?


TheLudlows:
The description in the document is 5M https://docs.influxdata.com/influxdb/v1.7/concepts/tsi-details/
config file ` max-index-log-file-size = "1m"` and in config.go is  `DefaultMaxIndexLogFileSize = 1 * 1024 * 1024`. So what'smax index log file size?

russorat:
@TheLudlows thanks for opening this. Looks like the default was changed in this PR: https://github.com/influxdata/influxdb/commit/92d38414f2eb51753e53797ff2294f022eaa4305 and we need to update the docs.

PR #2669, opened 2020-02-25
Update TSI max log file size

pierwill:


russorat:
Would it be useful to mention that the size can be configured and link to the config option?

pierwill:
Probably yes, I'd say. I'll look into this— 👍

russorat:
closes #2668

PR #2685, opened 2020-03-04
Update replication factor info

kelseiv:
Published changes to address incorrect replication factor info throughout the docs. 

Created [separate issue](https://github.com/influxdata/docs.influxdata.com/issues/2713) to address procedures/scenarios in _Rebalancing Enterprise Clusters_.
................................................................................................................................

Update doc to clarify relationship between replication factor and # of data nodes.

Include the following info where applicable:

- The number of data nodes in a cluster **must be evenly divisible** by the replication factor. For example, a replication factor of 2 works with 2, 4, or 6 data nodes, and so on. A replication factor of 3 works with 3, 6, or 9 data nodes, and so on.

-  If the number of data nodes isn't evenly divisible by the replication factor, data may be distributed unevenly across the cluster, causing poor query performance.

- To store more series within the database, increase the number of data nodes and update the replication factor as needed. For example, if you have RF=3 in a 6 node cluster, you can add 3 nodes or change the RF to 2 and add 2 data nodes. 

- Decreasing the replication factor (fewer copies of data in a cluster) may reduce query performance.


content/enterprise_influxdb/v1.7/concepts/glossary.md
```diff
@@ -53,19 +53,15 @@ Related entries: [meta node](#meta-node)
 
 The attribute of the retention policy that determines how many copies of the
 data are stored in the cluster.
-InfluxDB replicates data across `N` data nodes, where `N` is the replication
-factor.
+A cluster fully replicates data across each data node in the cluster.
```
timhallinflux: if and only if, the replication factor is equal to the number of data nodes.  If the number of data nodes is greater than the replication factor, this is called under replicated clusters.  Meaning, each data node only has a portion of all the series.   (I'm not sure this is the best language....).

content/enterprise_influxdb/v1.7/concepts/glossary.md
```diff
@@ -53,19 +53,15 @@ Related entries: [meta node](#meta-node)
...
+Replicating copies ensures that data is available when a data node (or more) is unavailable.
```
timhallinflux: Replicating data ensures that more copies of the data is available for query.  It also beneficial in the event data nodes become unavailable for any reason.  This fulfills the promise of high availability.

content/enterprise_influxdb/v1.7/concepts/glossary.md
```diff
@@ -53,19 +53,15 @@ Related entries: [meta node](#meta-node)
...
 
-To maintain data availability for queries, the replication factor should be less
-than or equal to the number of data nodes in the cluster:
+The number of data nodes in a cluster **must be evenly divisible** by the replication factor. For example, a replication factor of 2 works with 2, 4, 6, or 8 data nodes, and so on. A replication factor of 3 works with 3, 6, or 9 data nodes, and so on. Any time you update the number of data nodes, you must update the replication factor accordingly. For example, if you have RF=3 in a 6 node cluster, you can add 3 nodes or change the RF to 2 and add 2 data nodes.
```
timhallinflux: Starting with [Any time you update]...  If you wish to increase the read or write capacity of a cluster, you should do so by adding a number of nodes which is also a multiple of the replication factor.  For example, if you wish to increase the capacity of a 6 node cluster with an RF=3, you should add 3 additional nodes.   If you wish to further increase the capacity continue to add nodes in groups of 3.

content/enterprise_influxdb/v1.7/concepts/glossary.md
sanderson: It doesn't "fully" replicate it on "each" node. If replication factor is set to 2, it ensures that each series is stored on 2 separate nodes. The only time data will be **fully** replicated on **each** node in the cluster is if the replication factor is equal to the number of data nodes.

content/enterprise_influxdb/v1.7/concepts/glossary.md
```diff
@@ -53,19 +53,15 @@ Related entries: [meta node](#meta-node)
...
 
-* Data is fully available when the replication factor is greater than the
-number of unavailable data nodes.
-* Data may be unavailable when the replication factor is less than the number of
-unavailable data nodes.
+> **Important:** If the replication factor isn't evenly divisible into the number of data nodes, data may be distributed unevenly across the cluster and cause poor query performance. Likewise, decreasing the replication factor (fewer copies of data in a cluster) may reduce query performance depending on query load.
```
timhallinflux: poor performance in general...not just query.  Write performance may also suffer.

content/enterprise_influxdb/v1.7/concepts/glossary.md
sanderson: Replication ~copies~ ensures ~that~ data is available when...

content/enterprise_influxdb/v1.7/concepts/glossary.md
sanderson: - Include "by the replication factor" in the bold text.
- Any time you update the number of data nodes, ~you must~ update the...

content/enterprise_influxdb/v1.7/guides/rebalance.md
```diff
@@ -46,31 +45,27 @@ Performing a rebalance while writing historical data can lead to data loss.
...
 
-For demonstration purposes, the next steps assume that you added a third
+For demonstration purposes, this procedures shows how to add a third
 data node to a previously two-data-node cluster that has a
 [replication factor](/influxdb/v1.7/concepts/glossary/#replication-factor) of
-two.
-This rebalance procedure is applicable for different cluster sizes and
-replication factors, but some of the specific, user-provided values will depend
-on that cluster size.
+two; alter the replication factor and data nodes as needed for your for your cluster.
```
sanderson: "for your" is repeated.

content/enterprise_influxdb/v1.7/guides/rebalance.md
```diff
@@ -46,31 +45,27 @@ Performing a rebalance while writing historical data can lead to data loss.
 
 ## Rebalance Procedure 1: Rebalance a cluster to create space
 
-For demonstration purposes, the next steps assume that you added a third
+For demonstration purposes, this procedures shows how to add a third
```
timhallinflux: why is this demonstration purposes?

You are describing how to:
a) increase the query capacity of a cluster by increasing the number of replicas. on a 2 node cluster...going from RF=2 to RF=3 AND adding a data node.
b) alternatively, you can increase the capacity by under replicating the data...meaning going from full replication (DN=2 and RF=2) to (DN=4 and keeping RF=2).

content/influxdb/v1.7/concepts/glossary.md
```diff
@@ -159,10 +159,11 @@ See [Data Exploration](/influxdb/v1.7/query_language/data_exploration/), [Schema
 
 ## replication factor  
 
-The attribute of the retention policy that determines how many copies of data to concurrently store (or retain) in the cluster. Replicating copies ensures that data is available when a data node (or more) is unavailable.
+The attribute of the retention policy that determines how many copies of data to concurrently store (or retain) in the cluster. A cluster fully replicates data across each data node in the cluster. Replicating copies ensures that data is available when a data node (or more) is unavailable.
```
sanderson: Same comment as above.

content/enterprise_influxdb/v1.7/guides/rebalance.md
```diff
@@ -247,14 +238,12 @@ size on the original data nodes and increased the cluster's write throughput.
 For demonstration purposes, the next steps assume that you added a third
```
timhallinflux: why is this "demonstration purposes"?

content/influxdb/v1.7/query_language/database_management.md
```diff
@@ -257,11 +257,11 @@ duration is `INF`.
 - The `REPLICATION` clause determines how many independent copies of each point
 are stored in the [cluster](/influxdb/v1.7/high_availability/clusters/).
 
-- By default, the replication factor `n` usually equals the number of data nodes. However, if you have four or more data nodes, the default replication factor `n` is 3.
+- A cluster fully replicates data across each data node in the cluster. This lets a data node respond to a query without communicating with other data nodes.
```
sanderson: Same as above. The data node still needs to communicate with other nodes to respond to a query, but the replication ensures the data it needs is on more than one server and the chances of that data being unavailable go down.

content/enterprise_influxdb/v1.7/guides/rebalance.md
timhallinflux: hmmm....maybe we need to revisit these scenarios as a whole?  "Creating space" means simply adding disk space....not sure I would change the data nodes or replication factor for this use case.

content/influxdb/v1.7/guides/hardware_sizing.md
```diff
@@ -90,9 +90,9 @@ The InfluxDB Enterprise web server is primarily an HTTP server with similar load
 
 ### Data nodes
 
-A cluster with one data node is valid but has no data redundancy. Redundancy is set by the [replication factor](/influxdb/v1.7/concepts/glossary/#replication-factor) on the retention policy the data is written to. Where `n` is the replication factor, a cluster can lose `n - 1` data nodes and return complete query results.
+A cluster with one data node is valid but has no data redundancy. Redundancy is set by the [replication factor](/influxdb/v1.7/concepts/glossary/#replication-factor) on the retention policy the data is written to. To ensure data is successfully replicated across a cluster, the number of data nodes in a cluster **must be evenly divisible** by the replication factor.
 
->**Note:** For optimal data distribution within the cluster, use an even number of data nodes.
+>**Note:** If data nodes aren’t divisible by the replication factor, data may be distributed unevenly across the cluster, causing poor query performance.
```
timhallinflux: remove query.

content/influxdb/v1.7/query_language/database_management.md
timhallinflux: we need to describe full replication and partial/under replication scenarios....

content/influxdb/v1.7/query_language/database_management.md
```diff
@@ -303,6 +303,10 @@ This setting is optional.
 The query creates a retention policy called `one_day_only` for the database
 `NOAA_water_database` with a one day duration and a replication factor of one.
 
+> **Note:** To ensure data is successfully replicated across a cluster, the number of data nodes in a cluster **must be evenly divisible** by the replication factor.
+
+A cluster fully replicates data across each data node in the cluster. This lets a data node respond to a query without communicating with other data nodes. If data nodes aren’t divisible by the replication factor, data may be distributed unevenly across the cluster, causing poor query performance.
```
sanderson: Same as above.

content/enterprise_influxdb/v1.7/guides/rebalance.md
timhallinflux: perhaps engaging Joe from support to identify which scenarios we really want to document and advocate for here would be wise?

sanderson:
@timhallinflux, @kelseiv Thoughts on language – "under replicated" makes it sound like it's a bad thing. I think "paritally replicated" would be better. It's not as fault tolerant as full replication, but it's still a viable replication strategy.

- full replication vs partial replication
- fully replicated vs partially replicated

content/enterprise_influxdb/v1.7/guides/rebalance.md
kelseiv: think intention was "for example purposes" ....we're assuming this cluster configuration, but your cluster may have a different number of nodes, so alter your numbers accordingly.

PR #2695, opened 2020-03-05
Edit line protocol reference for special characters

### LINKED ISSUE #2676: LineProtocol - tag value cannot end with "\" even if escaped
A tag value cannot end with a backslash, even if escaped. You will get the following error:
> 2020-03-02T15:09:50Z E! [outputs.influxdb] When writing to [http://127.0.0.1:8086]: received error unable to parse {LineProtocolStringhere} : invalid tag format

But a tag value can end with spaces and other special chars.

So for example:
``` 
#Those are invalid rows
test, Tag1=HelloThere\ value1=1
test, Tag1=HelloThere\\ value1=1

#This is valid
test, Tag1=HelloThere\  value1=1
``` 

I've not found any reference to this limitation in the docs (but I remember reading something like it somewhere in a comment)

###### URL for relevant page?
- https://docs.influxdata.com/influxdb/v1.7/write_protocols/line_protocol_reference/
- https://docs.influxdata.com/influxdb/v1.7/write_protocols/line_protocol_tutorial/
- Telegraf v1.13.2
- InfluxDB v1.7.8
- https://docs.influxdata.com/influxdb/v1.7/write_protocols/line_protocol_reference/
- https://docs.influxdata.com/influxdb/v1.7/write_protocols/line_protocol_tutorial/

pierwill:
Note that tag values cannot end with a backslash.

Closes #2676.

### DIFF
--- a/content/influxdb/v1.7/write_protocols/line_protocol_reference.md
+++ b/content/influxdb/v1.7/write_protocols/line_protocol_reference.md
@@ -195,13 +195,12 @@ measurement require both double quotes and escaped (`\`) double quotes in the
 
 ### Special characters
 
-You must use a backslash character `\` to escape the following special characters:
+In field values, you must use a backslash character (`\`) to escape:
 
-* In string field values, you must escape:
-  * double quotes  
-  * backslash character
+- double quotes (`"`)
+- the backslash character itself (`\`).
 
-For example, `\"` escapes double quote.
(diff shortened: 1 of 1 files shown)


theodiefenthal:
Why was that notice in docs removed in InfluxDB 1.8 docs? still having this issue with InfluxDB 1.8.4.

Issue #2667, opened 2020-02-24
Document v2.0 /write API compatibility


timhallinflux:
###### URL for relevant page?
https://docs.influxdata.com/influxdb/v1.7/tools/api/

###### What products and version are you using?
InfluxDB 1.8

InfluxDB 1.8 introduces compatibility for the 2.0 write API.  This enables users who are on 1.8 to adopt new v2.0 client libraries now.  With the arrival of 1.8, users should be able to leverage both Flux (as their primary query language) and the 2.0 write API making their transition to 2.0 OSS, Cloud and Enterprise seamless.

However, we should document the best practices to ensure that this is as easy as possible for our users.  

Tracking issue:
https://github.com/influxdata/influxdb/issues/16898

PR:
https://github.com/influxdata/influxdb/pull/16908/

pierwill:
Closed via #2701.

pierwill:
#2733

PR #2701, opened 2020-03-09
Document /api/v2/write API endpoint

pierwill:
- Add /v2/write/ HTTP endpoint to API docs
- Mention the new endpoint in client library docs.
- Link to version 2 client libraries.

Local build succeeds.


pierwill:
Closes #2667.

sanderson:
@timh, when using the `/v2/write` endpoint, does the `bucket` translate to the database and the `org` translate to the retention policy?

We likely need to add an example of using one of the 2.0 client libraries to write to a 1.8+ database.

content/influxdb/v1.8/tools/api.md
```diff
@@ -592,6 +593,15 @@ Content-Length: 33
 {"error":"authorization failed"}
 ```
 
+## `/v2/write/` HTTP endpoint
```
ayang64: i think this should be `/api/v2/write`.

PR #2735, opened 2020-03-23
Update api.md

hrbonz:
The chunked option for the line protocol is split in two parameters. The doc conflates the two in one single parameter (which might have been the case in the past) but sniffing the traffic on chunked queries showed that the current behavior should be the one proposed here.


pierwill:
Thanks, @hrbonz !

hrbonz:
No problem, I found that out when doing testing for this PR https://github.com/influxdata/influxdb-python/pull/753, would appreciate a **nudge** to get this moving ;)

Issue #2694, opened 2020-03-05
Wrong user/permissions on private cert


sanderson:
Users have run into an issue where we direct them to set the wrong permissions on the certificate private key files, so the `influxdb` user cannot read them.

```
sudo chmod 600 /etc/ssl/<private-key-file>
```

~Needs to be updated to:~

[ deleted ]

**Update:** We need to make it clear that the user running InfluxDB must own the certificate files. In general, InfluxDB shouldn't be run by `root`, so certificate files should not be owned by the `root` user.

###### URL for relevant page?
- https://docs.influxdata.com/enterprise_influxdb/v1.7/guides/https_setup/#setup-https-in-an-influxdb-enterprise-cluster
- https://docs.influxdata.com/influxdb/v1.7/administration/https_setup/#configure-influxdb-to-use-tls
- _There may be others_

sanderson:
`600` is the correct permissions. `644` is not secure enough for a private key. The issue is that the private key must be owned by the user running InfluxDB.

sanderson:
The enterprise docs say that the `root` user should own the certificate files. This needs to be updated. Whatever user is running InfluxDB needs to own the files. We generally do not recommend running InfluxDB as `root`.

pierwill:
Closed in #2740.

PR #2740, opened 2020-03-25
Fix user in Enterprise HTTPS docs

pierwill:


pierwill:
Closes #2694

sanderson:
Be sure to port this change to the `influxdb-enterprise-1.8` branch.

PR #2741, opened 2020-03-25
Rebalancing procedures

kelseiv:
- Update procedures for "Rebalancing InfluxDB Enterprise clusters" (https://docs.influxdata.com/enterprise_influxdb/v1.7/guides/rebalance)
- Eliminate duplicate data (truncate hot shards, identify cold shards, copy cold shards, confirm rebalance)
- Add section for "Why rebalance?" to cover cases when folks should rebalance
- Add section to describe "Full versus partial replication"
- Add numbered steps at beginning of the "Rebalance your cluster" procedure


timhallinflux:
@Joe-Influx and @andrew-influx -- sorry to bug guys, but we could really use a review of this updated content.  I'd like to get this published.

kelseiv:
> @Joe-Influx and @andrew-influx -- sorry to bug guys, but we could really use a review of this updated content. I'd like to get this published.

Thank you, @timhallinflux - @Joe-Influx reached out this morning and is taking another look at this one.  I should have added a note here!

kelseiv:
> A few notes --
> 
> Line 42 -- Truncate shards in this context creates a possible assumption that the shard will be "cold". Could be worth another warning or a note that AE may be advisable or revisit in 48 hours as a followup to any procedure. (Worst danger is Entropy so maybe AE??). There is a dilemma here that completing a full rebalance in one sweep in 4-6 hours may have a few exceptions to any rule. May be something they have to iterate through at least twice.
> 
> Line 141 -- This is a bit of overkill. In theory Influx should give a failure if the shard copy operation is not successful. In documentation I think we balance best case with a few warnings but again, the danger is entropy or data loss. For this line the worst case would be entropy, which AE could fix in an ideal scenario.
> 
> Line 196 -- Use TCPName / Data node id as terms not TCP address. Meta nodes track TCPNames and Data node id's, which then map to DNS or IP's as needed.

Thank you, @andrew-influx! Will you make sure [this](https://github.com/influxdata/docs.influxdata.com/pull/2741/commits/96ae401ce6d8ace1cc56797da077a9d806b23615) addresses all the issues you surfaced?

PR #2756, opened 2020-04-09
Update Enterprise TLS certificate docs

### LINKED ISSUE #2681: In Section 4 have the cert and key name for meta nodes different than influxdb example
###### URL for relevant page?
https://docs.influxdata.com/enterprise_influxdb/v1.7/guides/https_setup/

Influx 1.7.10 Ent

Talked to Support to confirm. Andrew Spencer

We should call the crt and key different names for the InfluxDB side and Meta side. Right now both are labled as influxdb.crt and influxdb.key. So customers are trying to use the same cert for each. I would recommend influxdb-meta.crt and influxdb-meta.key on the meta section picture.

image

pierwill:
Now suggests the possibility of using multiple certificates for data and meta nodes.

Fixes #2681.

### DIFF
--- a/content/enterprise_influxdb/v1.7/guides/https_setup.md
+++ b/content/enterprise_influxdb/v1.7/guides/https_setup.md
@@ -55,23 +55,23 @@ that combine the private key file and the signed certificate file into a single 
 
     If using a certificate provided by a CA, follow their instructions to download the certificate files.
 
-    If using a self-signed certificate, use the `openssl` utility to create a certificate.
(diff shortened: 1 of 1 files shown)


pierwill:
I've simplified the self-signed certificate generation section and moved it into a note, since best practices for Enterprise deployments is to use signed certificates. This also makes the guide a bit more generic to allow for multiple sets of certificates. This revision gives an example of one each for meta and data nodes.

PR #2788, opened 2020-04-20
add release notes for Kapacitor 1.5.5

kelseiv:
Add Kapacitor 1.5.5 release notes

### DIFF
--- a/content/kapacitor/v1.5/about_the_project/releasenotes-changelog.md
+++ b/content/kapacitor/v1.5/about_the_project/releasenotes-changelog.md
@@ -5,6 +5,16 @@ menu:
     parent: about_the_project
 ---
 
+## v1.5.5 [2020-04-20]
+
+## Features
+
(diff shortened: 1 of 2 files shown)


docmerlin:
@kelseiv that is a good call on the wording.

kelseiv:
Thanks, @timhallinflux - not in install or config docs, but did see

 MD5 in  [_Calculating rates across joined series + backfill_](https://docs.influxdata.com/kapacitor/v1.5/guides/join_backfill/#the-data-generator).

Does it make sense to drop md5 link here and/or update [page.zip](https://docs.influxdata.com/downloads/pages.zip)?

docmerlin:
@kelseiv I'm not tim, but since you asked me to comment.  If we are standardizing on sha256 everywhere,  then it makes sense to drop the md5 link.

kelseiv:
thank you, @docmerlin! Do we need to update /repackage any of the python script in page.zip?

docmerlin:
@kelseiv no

Issue #2802, opened 2020-04-28
Document running Telegraf using remote configuration files


sanderson:
When starting Telegraf, you can use a URL with the `-config` flag to pull a configuration file from a remote endpoint and start Telegraf using that configuration.

#### Related links
- https://docs.influxdata.com/telegraf/v1.14/administration/configuration/
- https://github.com/influxdata/telegraf#run-telegraf-with-all-plugins-defined-in-config-file

kelseiv:
thanks, @sanderson, @rbetts! addressed in #2802

PR #2805, opened 2020-04-29
update for #2802

kelseiv:
minor edits and update to include remote URL endpoint option for `-config` flag

### DIFF
--- a/content/telegraf/v1.14/administration/configuration.md
+++ b/content/telegraf/v1.14/administration/configuration.md
@@ -19,20 +19,20 @@ telegraf config > telegraf.conf
 ```
 
 To generate a configuration file with specific inputs and outputs, you can use the
-`-input-filter` and `-output-filter` flags:
+`--input-filter` and `-output-filter` flags:
 
 ```
-telegraf -input-filter cpu:mem:net:swap -output-filter influxdb:kafka config
+telegraf --input-filter cpu:mem:net:swap -output-filter influxdb:kafka config
 ```
 
 ## Configuration file locations
 
-Use the `-config` flag to specify the configuration file location:
+Use the `--config` flag to specify the configuration file location:
 
-- Filename and path, for example: `-config /etc/default/telegraf`
-- Remote URL endpoint, for example: `-config http://remote-URL-endpoint`
+- Filename and path, for example: `--config /etc/default/telegraf`
+- Remote URL endpoint, for example: `--config "http://remote-URL-endpoint"`
 
-Use the `-config-directory` flag to include files ending with `.conf` in the specified directory in the Telegraf
+Use the `--config-directory` flag to include files ending with `.conf` in the specified directory in the Telegraf
 configuration.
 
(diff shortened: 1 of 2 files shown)


content/telegraf/v1.14/administration/configuration.md
```diff
@@ -19,19 +19,20 @@ telegraf config > telegraf.conf
...
 ## Configuration file locations
 
-The location of the configuration file can be set via the `--config` command
-line flag.
+Use the `-config` flag to specify the configuration file location:
 
-When the `--config-directory` command line flag is used, files ending with
-`.conf` in the specified directory will also be included in the Telegraf
+- Filename and path, for example: `-config /etc/default/telegraf`
+- Remote URL endpoint, for example: `-config http://remote-URL-endpoint`
```
kelseiv, line 33: need quotes around URL?

content/telegraf/v1.14/administration/configuration.md
```diff
@@ -19,19 +19,20 @@ telegraf config > telegraf.conf
 ```
 
 To generate a configuration file with specific inputs and outputs, you can use the
-`--input-filter` and `--output-filter` flags:
+`-input-filter` and `-output-filter` flags:
```
danielnelson, line 22: As an aside, I prefer using GNU style option format like `--input-filter` over `-input-filter`, even though Go doesn't require the extra dash.  This style is very common in Linux, and also I like to keep open the option of switching to a more powerful argument parsing library that uses the GNU style arguments and allows for short options.

content/telegraf/v1.14/administration/configuration.md
danielnelson: Depends on the exact URL and the shell in use, but I'd recommend showing it quoted

content/telegraf/v1.14/introduction/getting-started.md
```diff
@@ -10,39 +10,46 @@ menu:
     parent: Introduction
 ---
 
-## Getting started with Telegraf
-Telegraf is an agent written in Go for collecting metrics and writing them into InfluxDB or other possible outputs.
-This guide will get you up and running with Telegraf.
-It walks you through the download, installation, and configuration processes, and it shows how to use Telegraf to get data into InfluxDB.
+Use Telegraf (an agent written in Go) to collect and write metrics into InfluxDB or other supported outputs.
```
danielnelson, line 13: I'd love to come up with a better description of Telegraf.   I think mentioning Go so prominently isn't needed and doesn't make sense, maybe we should drop it.

content/telegraf/v1.14/administration/configuration.md
kelseiv: Thank you, @danielnelson! I appreciate the GNU standard. Would like to keep same standard across our docs, and will discuss w the docs team tomorrow.

content/telegraf/v1.14/introduction/getting-started.md
kelseiv: Tried hiding in parens, but dropping....even better!

PR #2809, opened 2020-04-30
Add word missing from sideload commands

AlexHeylin:


kelseiv:
Thank you, @AlexHeylin!

PR #2856, opened 2020-05-27
add info about OFFSET and WHERE in subqueries

kelseiv:
Closes https://github.com/influxdata/DAR/issues/90

Hi @ayang64 @jsteenb2 I was curious whether we should remove this note--or does this note still apply for "main" queries...just not valid for subqueries?

### DIFF
--- a/content/influxdb/v1.8/query_language/explore-data.md
+++ b/content/influxdb/v1.8/query_language/explore-data.md
@@ -707,7 +707,7 @@ SELECT_clause FROM_clause [WHERE_clause] GROUP BY [* | <tag_key>[,<tag_key]]
 The order of the [tag keys](/influxdb/v1.8/concepts/glossary/#tag-key) is irrelevant.
 
 If the query includes a [`WHERE` clause](#the-where-clause) the `GROUP BY`
-clause must appear after the `WHERE` clause.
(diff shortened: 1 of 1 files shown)


content/influxdb/v1.8/query_language/explore-data.md
```diff
@@ -707,7 +707,7 @@ SELECT_clause FROM_clause [WHERE_clause] GROUP BY [* | <tag_key>[,<tag_key]]
 The order of the [tag keys](/influxdb/v1.8/concepts/glossary/#tag-key) is irrelevant.
 
 If the query includes a [`WHERE` clause](#the-where-clause) the `GROUP BY`
-clause must appear after the `WHERE` clause.
```
kelseiv, line 710: @ayang64 @jsteenb2 
per the workaround in [EAR 1387](https://github.com/influxdata/EAR/issues/1387)
it appears we can include the WHERE clause after the GROUP BY clause. 

My question is this:
Do we remove this note altogether, or is this only the case when you have subqueries and still applies to "main" queries?
Is this "never" true or only true for "main" queries?

kelseiv:
moved to [v2 repo](https://github.com/influxdata/docs-v2/pull/1307/files)

Issue #2857, opened 2020-05-28
typo in telegraf concepts docs


jiteshk23:
###### URL for relevant page?
https://docs.influxdata.com/telegraf/v1.14/concepts/aggregator_processor_plugins/

###### What products and version are you using?
Just reading the docs

###### Where did you look before opening the issue?
Nowhere

NOTE Since aggregator plugins only aggregate metrics within their periods, historical data is not supported. In other words, if your metric timestamp is **more** than now() - period in the past, it will not be aggregated. 

Shouldn't it be **less** in the above note, or may be I misunderstood, in which case may be it can be made a bit more clear.

kelseiv:
Thank you @jiteshk23 - update to clarify that historical data is **earlier than** `now() - period`). Also, updated note to reflect that historical data can now be included in aggregated data.

PR #2858, opened 2020-05-29
Update to clarify historical data period

kelseiv:
Update note to clarify historical data period


kelseiv:
closes #2857

danielnelson:
Historical data isn't actually supported, only current data based on the wall clock.

danielnelson:
Can we also open this against the source documentation at the Telegraf repo:
https://github.com/influxdata/telegraf/blob/master/docs/AGGREGATORS_AND_PROCESSORS.md

PR #2860, opened 2020-06-02
Update Replacing Enterprise nodes procedures

kelseiv:
- Updated Replacing Enterprise meta and data node procedures in 1.7 and 1.8
- Added note that AE is disabled by default 1.7.2 and later
- Added tip about using copy-shard to manually copy shards, for example if folks run into "unexpected" shard issues when replacing a data node. Think it'd be helpful to add more detail --specifically what was unexpected re: shard issues that popped up in [DAR](https://github.com/influxdata/DAR/issues/91).

### DIFF
--- a/content/enterprise_influxdb/v1.7/administration/anti-entropy.md
+++ b/content/enterprise_influxdb/v1.7/administration/anti-entropy.md
@@ -10,7 +10,7 @@ menu:
 ---
 
 {{% warn %}}
-Prior to InfluxDB Enterprise 1.7.2, the Anti-Entropy (AE) service was enabled by default. When shards create digests with lots of time ranges (10s of thousands), some customers have experienced significant performance issues, including CPU usage spikes. If your shards include a small number of time ranges (most have 1 to 10, some have up to several hundreds) and you can benefit from the AE service, enable AE and monitor it closely to see if your performance is adversely impacted.
+Prior to InfluxDB Enterprise 1.7.2, the Anti-Entropy (AE) service was enabled by default. When shards create digests with lots of time ranges (10s of thousands), some customers have experienced significant performance issues, including CPU usage spikes. For versions 1.7.2 and later, AE is disabled by default. If your shards include a small number of time ranges (most have 1 to 10, some have up to several hundreds) and you can benefit from the AE service, enable AE and monitor it closely to see if your performance is adversely impacted.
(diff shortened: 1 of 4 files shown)


andrew-influx:
Adding a review.

Apologies for delay. Just got a confirmation on some intuitions as a best way forward.

To the best of my knowledge AE does not currently, and is not supporting a use case for backfilling an entire missing shard. The AE service was added to allow for a better mechanism to fill in gaps in data across 2 or more shard owners, but not for a completely missing shard. Completely missing shards only happen with human or admin intervention. The best method to backfill a missing shard to another data node is a complete shard-copy.

Using the update-data functionality (per a conversation with Joe) looks to be a bit of a misinterpretation. It is an older part of the tool set primarily used to update a hostname is everything else on disk stays the same. Using the update-data command to add in a new node would be incorrect, it would be add-data, per one understanding of the use case.

If you use add-data for a new server, then you would rebalance shards appropriately. 

The other understanding if you use update-data to have the new node (with no data on disk) join the cluster via the meta nodes as if it has all the same shards as a previous system means queries and writes will land on the new data node immediately. writes could cause new shard creation and queries will be handled with no or very minimal data to respond to them. There is no protection for this because the meta node accepts that data node as the same node id and assumed (via use case) the information on disk is the same, which it is not.

So even at a small scale (say 50gb) with some additional instructions to watch over AE with manually trigger shard repair (we have some uncertainty on relying on automatic repair for AE at this time) you will still have a matter of minutes to a few hours where queries will effectively fail with incorrect information.

Using add-data and shard rebalancing (with RF 2 or above) means 1 data node will be able to land the new writes and handle queries until shard copy operations are complete. You would also want to use truncate shards and all of the same steps for shard rebalancing.

Last point that came up in discussions is a FR was looked at (per Joe) for a shard rebalancing feature but shard copies can be intensive and could take down data nodes if not monitored by an admin or human operator, so even a similar shard rebalancing feature is not something Influx currently does or will handle.

Summary -- Unfortunately data node replacement can't be done using the update-data and AE. The original understanding (assumption) was this something AE could do, and would do better in time (1.8.x forward), but neither is true, unfortunately.

Let me know if I can provide anything additional.

kelseiv:
moved to [v2 repo](https://github.com/influxdata/docs-v2/pull/1309)

PR #2873, opened 2020-06-09
Python client library

### LINKED ISSUE #2872: Remove "Use a Client Library" Section
https://docs.influxdata.com/influxdb/v1.8/tools/api_client_libraries/

https://docs.influxdata.com/influxdb/v1.8/tools/api_client_libraries/#use-a-client-library
this section is specific to python and doesn't really make sense on this page.

kelseiv:
closes #2872

Russ, I've clarified that the install instructions are for Python...our goal is to ultimately include install instructions for all client libraries.

Alternatively, we could hide the instructions for now (if it's more confusing to have a one-off). Thoughts?


russorat:
I don't see any files changed on this PR.

kelseiv:
I'm sorry, @russorat -- I updated numbering on master and checked [this change](https://github.com/influxdata/docs.influxdata.com/commit/22db0fc12f0491cd100b17e53d2dcae608ac3a38) in there...so here's what we have posted to the site [now](https://docs.influxdata.com/influxdb/v1.8/tools/api_client_libraries/#sidebar). Alternatively, we can hide the "Install and use the Python client library" procedure for now. Our goal is to add install/usage instructions for each library (the ones we're maintaining) and move them to separate pages.

PR #2880, opened 2020-06-17
Clarify conditional logic 

### LINKED ISSUE #2872: Remove "Use a Client Library" Section

### LINKED ISSUE #2878: flow of if statements is missing information about when an if statement will stop
###### URL for relevant page?
https://docs.influxdata.com/flux/v0.65/guides/conditional-logic/

Flux

Docs and Slack

The if statement will stop on the first match, and as the Flux team describes it "short circuit when possible".  For example, given this Flux:

`|> map(fn: (r) => ({
    // Retain all existing columns in the mapped row
    r with
    // Set the level column value based on the _value column
    level:
      if r._value >= 95.0000001 then "critical"
      else if r._value >= 85.0000001 "warning"
      else if r._value >= 70.0000001 then "high"
      else "normal"
    })`

Where r._value is 96, then the output would be "critical" because the if statement would match on the first line and then stop.

While this is possible to test and work out I think it should be explicitly mentioned in the docs.

kelseiv:
closes #2878

### DIFF
--- a/content/flux/v0.65/guides/conditional-logic.md
+++ b/content/flux/v0.65/guides/conditional-logic.md
@@ -4,7 +4,7 @@ seotitle: Query using conditional logic in Flux
 list_title: Conditional logic
 description: >
   This guide describes how to use Flux conditional expressions, such as `if`,
-  `else`, and `then`, to query and transform data.
+  `else`, and `then`, to query and transform data. **Flux evaluates statements from left to right and stops evaluating once a condition matches.**
 menu:
   flux_0_65:
     name: Conditional logic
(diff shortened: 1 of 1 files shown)


content/flux/v0.65/guides/conditional-logic.md
```diff
@@ -35,6 +35,21 @@ Conditional expressions are most useful in the following contexts:
...
+## Evaluating conditional expressions
+
+Flux evaluates statements in order and stops evaluating once a condition matches.
+
+For example, given the following statement:
+
+```js
+if r._value >= 95.0000001 and r._value <= 100.0 then "critical"
+else if r._value >= 85.0000001 and r._value <= 95.0 then "warning"
+else if r._value >= 70.0000001 and r._value <= 85.0 then "high"
```
sanderson, line 47: Very very very minor nit, but to avoid tiny gaps in coverage, use a combination of `>` and `>=`.

```js
if r._value > 95.0 and r._value <= 100.0 then "critical"
else if r._value > 85.0 and r._value <= 95.0 then "warning"
else if r._value > 70.0 and r._value <= 85.0 then "high"
else "normal"
```

sanderson:
These changes should be ported to docs-v2.

content/flux/v0.65/guides/conditional-logic.md
kelseiv: thank you, @sanderson - updated! Think you mean tiny overlaps?

sanderson: Yeah, you're right – overlaps. Sorry :)

PR #2895, opened 2020-06-26
misc memory updates

kelseiv:
addresses https://github.com/influxdata/DAR/issues/22


kelseiv:
moved to [v2 repo](https://github.com/influxdata/docs-v2/pull/1308)

PR #2911, opened 2020-07-13
Document configurable HTTP headers

### LINKED ISSUE #2910: Document configurable HTTP headers
PR: https://github.com/influxdata/influxdb/pull/18667

Users can now specify specific http headers for influxdb to return.

These pages should be updated:

- https://github.com/influxdata/docs.influxdata.com/blob/master/content/influxdb/v1.8/administration/config.md#http-endpoints-settings
- https://github.com/influxdata/docs.influxdata.com/blob/master/content/influxdb/v1.8/administration/https_setup.md and/or https://github.com/influxdata/docs.influxdata.com/blob/master/content/influxdb/v1.8/administration/security.md, noting that this can be used to help comply with security policies regarding headers.

pierwill:
Closes #2910

### DIFF
--- a/content/influxdb/v1.8/administration/config.md
+++ b/content/influxdb/v1.8/administration/config.md
@@ -682,6 +682,19 @@ This setting has no effect if [`auth-enabled`](#auth-enabled-false) is set to `f
 
 Environment variable: `INFLUXDB_HTTP_PING_AUTH_ENABLED`
 
+#### `http-headers`
+
+User-supplied [HTTP response headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers).
+Configure this section to return [security headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers#Security)
+such as `X-Frame-Options` or `Content Security Policy` where needed.
+
+Example:
+
+```toml
+[http.headers]
+  X-Frame-Options = "DENY"
+```
+
 #### `https-enabled = false`
 
 Determines whether HTTPS is enabled.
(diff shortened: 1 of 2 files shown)


content/influxdb/v1.8/administration/config.md
```diff
@@ -682,6 +682,19 @@ This setting has no effect if [`auth-enabled`](#auth-enabled-false) is set to `f
 
 Environment variable: `INFLUXDB_HTTP_PING_AUTH_ENABLED`
 
+#### `http-headers`
+
+User-supplied [HTTP response headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers).
+Configure this section to return [security headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers#Security)
```
kelseiv, line 688: @pierwill could folks also configure headers other than security here? For example, `content_encoding = "gzip"`?

pierwill: Definitely. Security headers is an important use case we want to highlight here.

pierwill: Maybe add a sentence in the middle here: "Users can configure any header..."

timhallinflux: I think the question is whether InfluxDB can respond to the header.... in the case of gzip that is a kind of behavior change....

content/influxdb/v1.8/administration/security.md
```diff
@@ -36,6 +36,14 @@ HTTPS can also verify the authenticity of the InfluxDB server to connecting clie
 Resources:
 [Enabling HTTPS](/influxdb/v1.8/administration/https_setup/)
 
+## Configure security headers
+
+HTTP headers allow servers and clients to pass additional information along with requests.
+Certain headers help [enforce security](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers#Security) properties.
+
+Resources:
+[Configure HTTP headers](http://localhost:1314/influxdb/v1.8/administration/config/#http-headers)
```
timhallinflux, line 45: I think this URL looks wrong...  can you review/address?

pierwill: Yep, fixed!

Issue #2904, opened 2020-07-09
Add additional path for OSS to Enterprise migration


noramullen1:
###### URL for relevant page?

https://docs.influxdata.com/enterprise_influxdb/v1.8/guides/migration/

@timhallinflux 's  notes:

Did some digging and the team does not typically use this approach.  Support ran into issues that were reported and fixed.  However they typically preach another method that is a best practice for large data sets and zero downtime.  We should have both documented as suggested approaches but for different scenarios.  The approach typically recommended by support involves:
1. Set up the cluster
2. Take a portable backup from OSS
3. Restore on the cluster
4. Dual write
5. Export the data from the time the backup was taken to when the dual write started
6. Import the data

This gives them 100% uptime with a fresh replicated cluster and the ability to do a full test and acceptance cycle

kelseiv:
Migrate OSS to InfluxDB Enterprise

Currently documented:

1. Upgrade InfluxDB to the latest version
2. Set up InfluxDB Enterprise meta nodes
3. Set up InfluxDB Enterprise data nodes
4. Upgrade the InfluxDB binary on your OSS instance
5. Add the upgraded OSS instance to the InfluxDB Enterprise cluster
6. Add existing data nodes back to the cluster
7. Rebalance the cluster

- if Support is not using this currently documented process, do we want to keep it in docs/is there a reason or benefit to migrating this way?

**The following migration process is recommended for large data sets and zero downtime.**

To migrate from OSS to Enterprise with no down time and complete a full test and acceptance cycle (do folks not have the ability to do this test/acceptace cycle with the current process?). 

Do both migration paths have the same prerequisites (folks still need 1.7.10 or later to use the portable backup and we still want them to upgrade OSS and Enterprise to the latest version)?

1. Set up the cluster (same as steps 2 and 3 above?)
2. Take a portable backup from OSS:
    ```sh
     influxd backup -portable -host <IP address>:8088 /tmp/mysnapshot
    ```
    For more information, see [`-backup`](/influxdb/v1.8/administration/backup_and_restore/#backup)
3. Restore on the cluster by running the following:
     ```sh
    influxd restore -portable  [ -host <host:port> ] <path-to-backup-files>
    ```
       For more information, see [`-restore`](/influxdb/v1.8/administration/backup_and_restore/#restore)
[...]

PR #2919, opened 2020-07-15
new migration process

kelseiv:
Closes https://github.com/influxdata/docs.influxdata.com/issues/2904

### DIFF
--- a/content/enterprise_influxdb/v1.8/guides/migration.md
+++ b/content/enterprise_influxdb/v1.8/guides/migration.md
@@ -24,7 +24,7 @@ Migrate a running instance of InfluxDB open source (OSS) to an InfluxDB Enterpri
 
 - Deletes data in existing InfluxDB Enterprise data nodes
 - Transfers all users from the OSS instance to the InfluxDB Enterprise cluster
-- Requires downtime for the OSS instance
+
 {{% /warn %}}
 
 ## Migrate to InfluxDB Enterprise
@@ -35,8 +35,15 @@ Complete the following tasks:
 2. [Set up InfluxDB Enterprise meta nodes](#set-up-influxdb-enterprise-meta-nodes)
 3. [Set up InfluxDB Enterprise data nodes](#set-up-influxdb-enterprise-data-nodes)
 4. Do one of the following:
-   - Migrate a large data set with zero downtime
-   - Migrate a smaller data set with downtime
+   - [Migrate a data set with zero downtime](#migrate-a-data-set-with-zero-downtime). This method lets you move data between OSS to Enterprise as you're testing the migration.
+   - [Migrate a data set with downtime](#migrate-a-data-set-with-downtime). This method is useful if you're not able to run a portable backup, for example, for any of the following reasons:
+     - Data set exceeds a certain size (roughly 10 databases or more, or more than 500 shards per database)
+     - Hardware requirements aren't available
+     - Time constraints
+
+     {{% note %}}
+     With this method, you cannot move data from Enterprise back to OSS.
+     {{% /note %}}
 
 ### Upgrade InfluxDB to the latest version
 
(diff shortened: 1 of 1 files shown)


content/enterprise_influxdb/v1.8/guides/migration.md
```diff
@@ -26,40 +27,46 @@ Migrate a running instance of InfluxDB open source (OSS) to an InfluxDB Enterpri
...
 1. [Upgrade InfluxDB to the latest version](#upgrade-influxdb-to-the-latest-version)
 2. [Set up InfluxDB Enterprise meta nodes](#set-up-influxdb-enterprise-meta-nodes)
 3. [Set up InfluxDB Enterprise data nodes](#set-up-influxdb-enterprise-data-nodes)
-4. [Upgrade the InfluxDB binary on your OSS instance](#upgrade-the-influxdb-oss-instance-to-influxdb-enterprise)
-5. [Add the upgraded OSS instance to the InfluxDB Enterprise cluster](#add-the-new-data-node-to-the-cluster)
-6. [Add existing data nodes back to the cluster](#add-existing-data-nodes-back-to-the-cluster)
-7. [Rebalance the cluster](#rebalance-the-cluster)
+4. Do one of the following:
+   - Migrate a large data set with zero downtime
+   - Migrate a smaller data set with downtime
```
kelseiv, line 39: want to avoid terms like "large" and "smaller" if possible...is there another way to put this? export/import method and...?

timhallinflux: perhaps we can characterize this across a couple of dimensions:
1) # of databases... > 10?
and/or
2) sum of the # of shards/database (typically driven by retention policy and shard duration)  > 500?

We define this upfront and then use that to drive the recommended path?

Let's confirm this characterization with support.

timhallinflux:
@gshif -- worth reviewing this approach. Particularly as we think about ease of migration from OSS 2 to Enterprise 1

content/enterprise_influxdb/v1.8/guides/migration.md
```diff
@@ -14,52 +14,67 @@ menu:
...
-4. [Upgrade the InfluxDB binary on your OSS instance](#upgrade-the-influxdb-oss-instance-to-influxdb-enterprise)
-5. [Add the upgraded OSS instance to the InfluxDB Enterprise cluster](#add-the-new-data-node-to-the-cluster)
-6. [Add existing data nodes back to the cluster](#add-existing-data-nodes-back-to-the-cluster)
-7. [Rebalance the cluster](#rebalance-the-cluster)
+4. Do one of the following:
+   - [Migrate a data set with zero downtime](#migrate-a-data-set-with-zero-downtime). This method lets you move data between OSS and Enterprise as you're testing the migration.
+   - [Migrate a data set with downtime](#migrate-a-data-set-with-downtime). With this method, you cannot move data from Enterprise back to OSS. This method is useful if you're not able to run a portable backup, for example, for any of the following reasons:
+     - Data set exceeds a certain size (roughly 10 databases or more, or more than 500 shards per database)
+     - Hardware requirements aren't available
+     - Time constraints
```
kelseiv: - Please confirm lines 44-47 are indeed true. 
- What add'l hardware is required (or how much)? 
- Does this method take more time? If yes, how much more? Does it make sense to say why?

content/enterprise_influxdb/v1.8/guides/migration.md
```diff
@@ -14,52 +14,67 @@ menu:
...
 1. [Upgrade InfluxDB to the latest version](#upgrade-influxdb-to-the-latest-version)
 2. [Set up InfluxDB Enterprise meta nodes](#set-up-influxdb-enterprise-meta-nodes)
 3. [Set up InfluxDB Enterprise data nodes](#set-up-influxdb-enterprise-data-nodes)
-4. [Upgrade the InfluxDB binary on your OSS instance](#upgrade-the-influxdb-oss-instance-to-influxdb-enterprise)
-5. [Add the upgraded OSS instance to the InfluxDB Enterprise cluster](#add-the-new-data-node-to-the-cluster)
-6. [Add existing data nodes back to the cluster](#add-existing-data-nodes-back-to-the-cluster)
-7. [Rebalance the cluster](#rebalance-the-cluster)
+4. Do one of the following:
+   - [Migrate a data set with zero downtime](#migrate-a-data-set-with-zero-downtime). This method lets you move data between OSS and Enterprise as you're testing the migration.
+   - [Migrate a data set with downtime](#migrate-a-data-set-with-downtime). With this method, you cannot move data from Enterprise back to OSS. This method is useful if you're not able to run a portable backup, for example, for any of the following reasons:
```
kelseiv: Specified new migration path as being "zero downtime"....rather than use for "large" data sets. It sounds like the original migration path may be best for "large" data sets?

content/enterprise_influxdb/v1.8/guides/migration.md
Joe-Influx: I agree on the terminology but behavior here is hard to define. The number of databases and retention policies should not be too much of a factor but rather their size. Large data sets on smaller instances are going to take a long time.

content/enterprise_influxdb/v1.8/guides/migration.md
Joe-Influx: For a generic Enterprise cluster you would need 5 (3 meta 2 data) instances. If you only had access to this number you would not be able to run the OSS in parallel

content/enterprise_influxdb/v1.8/guides/migration.md
kelseiv: Fine detail. Thank you for clarifying, @Joe-Influx

content/enterprise_influxdb/v1.8/guides/migration.md
kelseiv: @Joe-Influx, I took the number of databases out. Is the following accurate:

- [Migrate a data set with downtime](#migrate-a-data-set-with-downtime). With this method, you cannot move data from Enterprise back to OSS. This method is useful if you're not able to run a portable backup, for example, for any of the following reasons:
     - Data set exceeds a certain size (more than 500 shards per database)
     - Hardware requirements aren't available (fewer than 5 instances (3 meta 2 data)) 
     - Time constraints
 
You mentioned folks might not be able to do a portable backup if they had time constraints...does it make sense to add further detail regarding time?

kelseiv:
moved this PR to the [v2 repo](https://github.com/influxdata/docs-v2/pull/1305/files#diff-0a1d167d4f3675a3e77ed7f3c5616511R44-R47)

PR #2925, opened 2020-07-22
Telegraf 1.15 new plugins

noramullen1:
1.x docs version of plugins (exact same as changes from 2.x PR that you already approved!)

### DIFF
--- a/data/telegraf_plugins.yml
+++ b/data/telegraf_plugins.yml
@@ -1045,6 +1045,13 @@ input:
     introduced: 1.9.0
     tags: [linux, macos, windows, servers, web]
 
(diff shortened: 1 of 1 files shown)


data/telegraf_plugins.yml
```diff
@@ -1045,6 +1045,13 @@ input:
     introduced: 1.9.0
     tags: [linux, macos, windows, servers, web]
 
+- name: NGINX Stream STS
```
sanderson, line 1048: Line 1048 needs to be indented one level.

kelseiv:
obsolete being we're combining Telegraf plugin docs in new structure

PR #2933, opened 2020-07-30
Markdown fix.

mgolosova:
Hugo v0.37 fails to recognize bulleted list without blank line:
https://docs.influxdata.com/influxdb/v1.8/concepts/key_concepts/


pierwill:
Thanks for this, @mgolosova!

PR #2937, opened 2020-07-31
fixed int() TICKscript function description removing duration, resolves #2936

### LINKED ISSUE #2936: duration cannot be converted into int64
_TICKscript syntax_ documatation states that:

> `int()` - converts a string, float64, Boolean or **duration** type to an int64

But if I try to convert a duration into an integer, I get following error:

> ```
> Cannot call function "int" with args (30m0s: duration), available signatures are [(boolean), (string), (int), (float)].
> ```

So I guess **duration** should be removed from `int()` function documentation.

###### URL for relevant page?
https://docs.influxdata.com/kapacitor/v1.5/tick/syntax/#working-with-tags-fields-and-variables

Kapacitor v1.5

Open issues with `duration`, `int`, `int64` or `integer` keywords.

mamosk:
If you try to convert a **duration** into an **int64** running a Kapacitor TICKscript then you'll get an error.
The documentation states this is possible, but it's not.
Just corrected this in v1.5 documentation, not sure if it has to be corrected in v1.4 documentation too.

### DIFF
--- a/content/kapacitor/v1.5/tick/syntax.md
+++ b/content/kapacitor/v1.5/tick/syntax.md
@@ -519,7 +519,7 @@ For more specific information, see [Alert node](/kapacitor/v1.5/nodes/alert_node
 Within lambda expressions it is possible to use stateless conversion functions to convert values between types.
 
    * `bool()` - converts a string, int64 or float64 to Boolean.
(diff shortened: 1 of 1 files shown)


kelseiv:
Thank you, @mamosk! I rolled over to 1.4 as well

PR #2949, opened 2020-08-12
added compact-throughput

mprintz:
Matching the doc for ent. up with OSS. Adding the compact-throughput along with the default values.

### DIFF
--- a/content/enterprise_influxdb/v1.8/administration/config-data-nodes.md
+++ b/content/enterprise_influxdb/v1.8/administration/config-data-nodes.md
@@ -258,6 +258,19 @@ This setting does not apply to cache snapshotting.
 
 Environmental variable: `INFLUXDB_DATA_CACHE_MAX_CONCURRENT_COMPACTIONS`
 
+#### `compact-throughput = "48m"`
+
+The rate limit, in bytes per second, that we will allow TSM compactions to write to disk. The default value for this is '48 minutes'.
+Note that short bursts are allowed to happen at a possibly larger value, set by `compact-throughput-burst`.
+
+Environment variable: `INFLUXDB_DATA_COMPACT_THROUGHPUT`  
+
+#### `compact-throughput-burst = "48m"`
+
+The rate limit, in bytes per second, that we allow TSM compactions to write to disk. The default value for this is '48 minutes'.
+
+Environment variable: `INFLUXDB_DATA_COMPACT_THROUGHPUT_BURST`  
+
 #### `compact-full-write-cold-duration = "4h"`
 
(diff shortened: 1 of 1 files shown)


kelseiv:
Thank you, @mprintz! 
If remove "we will allow" does it change the meaning..sounds like customers can change this value?

that <del>we will allow</del> TSM compactions <del>to</del> write to disk.

content/enterprise_influxdb/v1.8/administration/config-data-nodes.md
```diff
@@ -258,6 +258,19 @@ This setting does not apply to cache snapshotting.
 
 Environmental variable: `INFLUXDB_DATA_CACHE_MAX_CONCURRENT_COMPACTIONS`
 
+#### `compact-throughput = "48m"`
+
+The rate limit, in bytes per second, that we will allow TSM compactions to write to disk. The default value for this is '48 minutes'.
```
sanderson, line 263: If it's a rate limit in bytes, it doesn't make sense that the default is 48 "minutes." Should this be 48 million? If so, why specify this with a string rather than an integer? Probably a question for the engineers.

sanderson: Looking at the [source code](https://github.com/influxdata/influxdb/blob/c476da2153f9e06362c1852b7b510620bd89e79e/tsdb/tsm1/engine.go#L191-L193), this gets casted to an integer at runtime.

sanderson: Update to:

```
The maximum number of bytes per seconds TSM compactions write to disk. Default is `"48m"` (48 million).
```

content/enterprise_influxdb/v1.8/administration/config-data-nodes.md
```diff
@@ -258,6 +258,19 @@ This setting does not apply to cache snapshotting.
...
+Note that short bursts are allowed to happen at a possibly larger value, set by `compact-throughput-burst`.
+
+Environment variable: `INFLUXDB_DATA_COMPACT_THROUGHPUT`  
+
+#### `compact-throughput-burst = "48m"`
+
+The rate limit, in bytes per second, that we allow TSM compactions to write to disk. The default value for this is '48 minutes'.
```
sanderson, line 270: Update to:

```
The maximum number of bytes per seconds TSM compactions write to disk during brief bursts. Default is `"48m"` (48 million).
```

kelseiv:
Hey, @mprintz  thoughts on comments above? 
We're moving all content in this v1 repository to the `docs-v2` repository...would love to get your thoughts before we transfer content.

PR #2958, opened 2020-08-26
Typo fix in  flux-vs-influxql.md

thisiswhereitype:
yeild to yield


kelseiv:
Thank you, @thisiswhereitype!

